PR #35, opened 2024-01-15
feat: basic telemetry

### LINKED ISSUE #23: Add telemetry
Add basic telemetry so we can tell how many users start trying out this extension.

deboer-tim:
Logs an event whenever users with telemetry enabled do a build or try to launch vfkit, with basic anonymous details. We need to see what data this produces and it will clearly need to change as we evolve the extension, but gives us a basic view for now.

Fixes #23.


cdrage:
LGTM! Merge when ready.

cdrage (approved):

PR #45, opened 2024-01-16
Update workable badges / icons

cdrage:
Update workable badges / icons

Since the upstream changes have changed paramters (example, we use
'label' instead of 'name'), I've updated bootc for it to work.

Depends on: https://github.com/containers/podman-desktop/pull/5557

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,6 +28,8 @@ const bootcImageBuilderContainerName = '-bootc-image-builder';
 const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest-1704948606';
 let diskImageBuildingName: string;
 
+const telemetryLogger = extensionApi.env.createTelemetryLogger();
+
 export async function activate(extensionContext: ExtensionContext): Promise<void> {
   bootc = new BootC(extensionContext);
   await bootc?.activate();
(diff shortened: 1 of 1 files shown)


cdrage:
@deboer-tim suggestions on colours? Right now it's bright green.

deboer-tim:
> @deboer-tim suggestions on colours? Right now it's bright green.

I don't think we have any official colors yet, so for now I'd just tone it down a bit. e.g. use one of our palette colors, or pick one of the Tailwind defaults that looks nice: https://tailwindcss.com/docs/customizing-colors#default-color-palette

package.json
```diff
@@ -53,18 +53,21 @@
...
+      "badges/image": [
         {
           "when": "ostree.bootable in imageLabelKeys",
           "badge": {
-            "name": "bootc",
-            "style": "bg-green-600"
+            "label": "bootc",
+            "color": {
+              "dark": "#32CD32",
+              "light": "#FF0000"
```
benoitf, line 69: if you prefer I think you can choose a class name for the color instead of #rgb

cdrage: Doesn't look like you can with the background values as tailwind doesn't pre-render them if they are passed in dynamically (looks like you can only show it if somewhere in the code we have the colour code... 

I just tried with bg-green-300, but bg-sky-300 works. It's odd.

I've gone ahead and changed it to a light-green for now though!


https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/50dc89d7-27c1-4d6d-865d-185edd42c89c

deboer-tim: It's not surprising that blue didn't work, since we have a restricted palette and blue is not in it:
https://github.com/containers/podman-desktop/blob/main/tailwind.config.cjs

No idea why green didn't work, but sky looks good. As Florent said it's probably better to go with that (or any other class name that works) over rgb.

cdrage: That was just a test :) but green doesn't work since we have it in our palette, but we do not call green anywhere, so tailwind doesn't "pre-render" it / bring it in when compiling. I can't find the issue, but I remember we had this problem before.

I know that "sky" works because we use that colour for other container providers.

I don't have a problem going with sky!

slemeur:
🚀

PR #54, opened 2024-01-17
Update workable badges / icons

cdrage:
Update workable badges / icons

Since the upstream changes have changed paramters (example, we use
'label' instead of 'name'), I've updated bootc for it to work.

Depends on: https://github.com/containers/podman-desktop/pull/5557


deboer-tim (approved):
LGTM

Issue #24, opened 2024-01-12
Initial Github, extension hygiene


deboer-tim:
Make sure we have a Github description, review the CI, create issue templates, etc.

deboer-tim:
Github description updated, issue and PR templates added, plus PR check. Trying to think if there's anything else we need initially, otherwise I'll close this.

deboer-tim:
Closing this since all the initial things are done; will open up other issues/PRs if anything else comes up.

PR #62, opened 2024-01-18
feat: add issue templates

deboer-tim:
Adds issue and PR templates. Almost identical to what we have in Podman Desktop, just removed one option that didn't apply, removed releases, and changed a reference or two from Podman Desktop to this extension.

Part of #24.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,59 @@
+name: Bug 🐞
+description: Report a bug report
+labels: [kind/bug 🐞]
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Before opening a bug report, please search for the behaviour in the existing issues.
+
+        ---
+
+        Thank you for taking the time to file a bug report. To address this bug as fast as possible, we need some information.
+
+  - type: textarea
+    id: bug-description
+    attributes:
+      label: Bug description
+      description: What happened?
+    validations:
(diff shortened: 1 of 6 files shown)


benoitf (commented):
IMHO you'll need to create/rename current labels to match the one in the templates

`bug` --> `kind/bug 🐞`

etc.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,59 @@
...
+      description: "Which operating system are you on? Please provide the version as well. If you are on a Mac, please specify Apple silicon or Intel."
+      placeholder: "macOS Ventura 13.4 (Arm), Windows 11"
+    validations:
+      required: true
+
+  - type: dropdown
+    id: version
+    attributes:
+      label: Version
+      description: What version of the software are you running?
```
benoitf, line 36: might be too generic ? which version of the extension or bootc extension ?

deboer-tim: Updated to 'bootable container extension'

deboer-tim:
> IMHO you'll need to create/rename current labels to match the one in the templates

Not sure why I thought these were populated - done now.

benoitf (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
benoitf: thanks

Issue #55, opened 2024-01-18
Logically separate files


deboer-tim:
When we started coding there was an initial bootc.ts and we didn't use it, then put commands and everything in extension.ts. We need to pull these apart into logical files, to avoid merge conflicts/stepping on each other's toes, and make testing per-file more obvious.

PR #63, opened 2024-01-18
chore: refactor vfkit

deboer-tim:
### What does this PR do?

Move vfkit launching to its own file. No changes to implementation.

### Screenshot / video of UI

N/A

Part of #55.

### How to test this PR?

Just moving existing function, in part so tests can be targeted and clearly separated too.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -37,38 +37,7 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
   extensionContext.subscriptions.push(
 
     extensionApi.commands.registerCommand('bootc.vfkit', async container => {
-      let imageLocation = container.labels['bootc.build.image.location'];
-
-      // Check that vfkit is installed and error if it isn't before executing
-      const telemetryData: Record<string, any> = {};
-      telemetryData.imageType = container.labels['bootc.build.type'];
-      try {
-        await extensionApi.process.exec('vfkit', ['--version']);
-      } catch (error) {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -37,38 +37,7 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
-      // if it is not raw or ami, we cannot launch with vfkit
-      if (container.labels['bootc.build.type'] !== 'ami' && container.labels['bootc.build.type'] !== 'raw') {
-        await extensionApi.window.showErrorMessage(`Unable to launch ${imageLocation} with vfkit: ${container.labels['bootc.build.type']} is not supported`);
-        telemetryData.error = 'unsupported-type';
-        telemetryLogger.logUsage('launchVfkit', telemetryData);
-        return;
-      }
-
-      await launchVfkit(imageLocation, telemetryData);
+      launchVFKit(container);
```
benoitf, line 40: I think you should await this call or catch the execution

probably need to add PR check about linter/formatter to this repository as well

benoitf (approved):

src/extension.ts
deboer-tim: Just added missing await for now, but likely needs better error handling in future.

Yes, big +1 for linter/formatter - I made this repo compliant last week but it already has a few issues again. I've fixed those in my next PR but needs to be part of PR check very soon.

PR #67, opened 2024-01-19
chore: add build to PR check

deboer-tim:
### What does this PR do?

Discussed with Charlie. We're not entirely sure what a build will catch over the existing lint/test, but since we're doing it post-merge it's better to know that it passed in the PR.

### Screenshot / video of UI

N/A

Part of #24.

### How to test this PR?

See PR status.


benoitf (changes requested):
I don't see why you're removing the build there ?

it's to check that the extension can be built on each platform. You may have people developing the extension on linux, macOS or Windows

The extension will run natively on these systems so I don't see the point of removing this check ?

.github/workflows/pr-check.yaml
```diff
@@ -75,3 +75,7 @@ jobs:
 
       - name: Run tests
         run: yarn test
+
+      - name: Run build
+        if: ${{ matrix.os=='ubuntu-22.04' }}
```
benoitf, line 80:
```suggestion
```

why do not build on each platform ?

you may have developers trying to build the extension on other platforms than Linux

.github/workflows/pr-check.yaml
```diff
@@ -75,3 +75,7 @@ jobs:
...
+        run: yarn build
```
benoitf, line 81: file is missing a EOF

deboer-tim: Added

.github/workflows/pr-check.yaml
deboer-tim: The goal for now was just to do everything from next-build, so when Windows build here failed with a generic and un-actionable message I reverted to only the platform next-build is on (ubuntu).

I just tried reverting that change and it built ok on windows twice, so hoping it was just a rare github/windows/yarn fluke and going with all three platforms.

PR #68, opened 2024-01-19
chore: cleanup bootc.ts

### LINKED ISSUE #55: Logically separate files

deboer-tim:
### What does this PR do?

The original template had a bootc.ts that we didn't use. None of the other extensions have this pattern, so we've decided to just clean it up for now. Tests merged in and extra files deleted.

### Screenshot / video of UI

N/A

Fixes #55.

### How to test this PR?

`yarn test`


benoitf (commented):
it's the pattern I try to push (some extensions have it)

it allows to have better/easier mocking tests without exporting functions everywhere,

extension.ts should only be empty

benoitf:
basically

```

  extensionContext.subscriptions.push(
    extensionApi.commands.registerCommand('bootc.vfkit', async container => {
      await launchVFKit(container);
    }),

    extensionApi.commands.registerCommand('bootc.image.build', async image => {
      await buildDiskImage(image);
    }),
  );
}
```

should go to bootc.ts

extension.ts should be minimalistic

benoitf:
and each time I've started an extension I've followed that approach but then all the people are adding everything inside extension.ts with imports from tons of different files 


maybe I should do the opposite: bootc.ts calling extension.ts class ?

cdrage (approved):

cdrage:
> and each time I've started an extension I've followed that approach but then all the people are adding everything inside extension.ts with imports from tons of different files
> 
> maybe I should do the opposite: bootc.ts calling extension.ts class ?

I think so. IMO extension.ts should be talking only to the extension api if that makes sense?

I'm good with this change though removing the `bootc.ts` for now until we have more robust / added tests.

deboer-tim:
> and each time I've started an extension I've followed that approach but then all the people are adding everything inside extension.ts with imports from tons of different files
> 
> maybe I should do the opposite: bootc.ts calling extension.ts class ?

Yeah, I think it was our natural behaviour to start putting things in extension.js because 'that's the entry point', and if we moved everything useful to bootc.ts then extension looks odd with just two functions? Might be one of those things that's better but doesn't feel natural to everyone. Or could just be learned behaviour too, since every extension I've seen uses extension.ts.

benoitf:
it's natural to proceed like that when you write good tests and use TDD

extension.js might looks odd but this is the worst testing case so it's to try to reduce to it's minimal the scope

benoitf (commented):
now you've almost remove all the existing tests

PR #70, opened 2024-01-20
feat: add erroring out / refactor / add tests

cdrage:
feat: add erroring out / refactor / add tests

### What does this PR do?

* Checks to see if the image has actually errored out before continuing
* Moves the "container utility" functions to a separate file
* Adds a lot more tests

### Screenshot / video of UI

N/A

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/38

### How to test this PR?

1. Start a build image
2. Stop the container during the build process.
3. It will fail as the container image will exit with error (1), rather
   than saying success

### DIFF
--- a/src/build-disk-image.ts
+++ b/src/build-disk-image.ts
@@ -21,10 +21,8 @@ import * as extensionApi from '@podman-desktop/api';
 import * as os from 'node:os';
 import * as fs from 'node:fs';
 import { resolve } from 'node:path';
-
-const bootcImageBuilderContainerName = '-bootc-image-builder';
-const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest-1704948606';
-let diskImageBuildingName: string;
+import * as containerUtils from './container-utils';
+import { bootcImageBuilderContainerName, bootcImageBuilderName } from './constants';
 
 const telemetryLogger = extensionApi.env.createTelemetryLogger();
 
@@ -90,10 +88,9 @@ export async function buildDiskImage(image: any) {
   return extensionApi.window.withProgress(
     { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + image.name },
     async progress => {
-      // Use the image name and append bootc-image-builder to the end
-      diskImageBuildingName = image.name.split('/').pop() + bootcImageBuilderContainerName;
-
+      const buildContainerName = image.name.split('/').pop() + bootcImageBuilderContainerName;
       let successful: boolean;
+      let logData: string = 'Build Image Log --------\n';
 
       // Create log folder
       if (!fs.existsSync(selectedFolder)) {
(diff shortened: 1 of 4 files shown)


cdrage:
In draft FYI! Will ping when in review.

@deboer-tim Letting you know in case this gets merged in and you have to rebase lots, this moves stuff to a different file + adds more tests.

cdrage:
Ready for review!

src/build-disk-image.ts
```diff
@@ -104,61 +103,68 @@ export async function buildDiskImage(image: any) {
...
-        await extensionApi.window.showErrorMessage(`Unable to build disk image: ${error}. Check logs at ${logPath}`);
       }
 
       // Mark the task as completed
       progress.report({ increment: -1 });
-
       telemetryLogger.logUsage('buildDiskImage', telemetryData);
 
-      // Only if success = true and type = ami
-      if ((successful && selectedType === 'ami') || selectedType === 'raw') {
```
cdrage, line 161: See if you can spot the logic error / hack for the demo haha. This part I fixed up since it kept saying success if you did raw...

src/container-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+  console.log('Waiting for container to exit: ', containerId);
+  let containerRunning = true;
+
+  while (containerRunning) {
+    await extensionApi.containerEngine.listContainers().then(containers => {
+      containers.forEach(container => {
+        if (container.Id === containerId && container.State === 'exited') {
+          // Let's stop the loop if the container has exited / stopped
+          containerRunning = false;
+          // Container.status reports "ex. Exited (1) Less than a second ago" when it
```
cdrage, line 85: Will now check if Exited (0) happens.

cdrage: You can trigger this by simply "stopping" the container mid-build.

cdrage:
~EDIT: Found a few more bugs~

Had an issue with removing the previous container build, since I was passing in the container name rather than ID, that's fixed now.

It's ready for review!

src/build-disk-image.ts
```diff
@@ -21,10 +21,10 @@ import * as extensionApi from '@podman-desktop/api';
 import * as os from 'node:os';
 import * as fs from 'node:fs';
 import { resolve } from 'node:path';
+import * as containerUtils from './container-utils';
 
 const bootcImageBuilderContainerName = '-bootc-image-builder';
-const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest-1704948606';
-let diskImageBuildingName: string;
+const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest';
```
deboer-tim, line 27: I don't think I'm comfortable with this yet - changing one argument could easily break our extension and they're likely not trying to maintain compatibility yet. Can we stick with a tagged version or sha, at least for now?

deboer-tim (approved):

cdrage:
@deboer-tim I've moved the constant variables to `constants.ts` so it's in a more unified place since we may be changing that section a lot as we update bootc-image-builder.

I've updated the image to the latest tag that has the `.iso` support / latest changes.

deboer-tim (approved):

src/build-disk-image.ts
benoitf: you might pin the release to a given sha but add an option that will be in preferences where as a user you might/could override or a checkbox 'dev/stable' or drop-down 'latest release/ dev release or extension's versions

So default is locked but you allow to test other versions

PR #72, opened 2024-01-22
chore: update provider

deboer-tim:
### What does this PR do?

I happened to notice that our extensions storage is in redhat.bootc-extension while minikube is in podman-desktop.minikube. Seems we forgot to update our provider when the repo was created in this org.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A


deboer-tim:
Decided not to do this for mow.

Issue #30, opened 2024-01-12
Store disk image info and add actions


deboer-tim:
Whenever we generate a disk image, update a json file in extension storage containing the image name, type, and disk image path.

Use this to add Launch VM or Open Folder actions on the original image.

deboer-tim:
We've added history, which for now is only used to pre-fill the output folder when generating again. We are dropping tentative plans for a vfkit launcher, so there is nothing else to do here and I'm closing this issue.

The rest of the image-generation history we're saving isn't required atm and could be removed... but since there's a good chance that we'll find other tools to launch with or have other actions in the future I plan to leave it in as it'll give us a good list of images pre-whatever-release that we can act on.

PR #73, opened 2024-01-22
feat: remember image build history

deboer-tim:
### What does this PR do?

Saves the image, type, and folder for every image build. For today, this just lets us remember the history of image build folders and default to the same folder the next time you build.

Why save more than the last folder, and why save the image name and type too? There is no registry of past disk images (like we have with container images, or containers), e.g. they aren't saved to an OCI registry. Although there's some initial discussion to do this, there no timeline.

Saving this data means we have a list of all past images and at least some options for providing actions on them, e.g. launching vfkit from the (container) image instead of just from the ephemeral container that was used to build it. Including the data in this PR means we don't need to deal with an imminent breaking format change.

### Screenshot / video of UI

N/A

Part of #30.

### How to test this PR?

Build an image, restart Podman Desktop, and confirm it remembers the output folder the next time you build.

### DIFF
--- a/src/build-disk-image.ts
+++ b/src/build-disk-image.ts
@@ -23,11 +23,12 @@ import * as fs from 'node:fs';
 import { resolve } from 'node:path';
 import * as containerUtils from './container-utils';
 import { bootcImageBuilderContainerName, bootcImageBuilderName } from './constants';
+import type { History } from './history';
 
 const telemetryLogger = extensionApi.env.createTelemetryLogger();
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export async function buildDiskImage(image: any) {
+export async function buildDiskImage(imageData: unknown, history: History) {
+  const image = imageData as { name: string; engineId: string };
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const telemetryData: Record<string, any> = {};
 
@@ -50,9 +51,10 @@ export async function buildDiskImage(image: any) {
   const selectedType = selection.format;
   telemetryData.imageType = selectedType;
 
+  const location = history.getLastLocation() || os.homedir();
   const selectedFolder = await extensionApi.window.showInputBox({
     prompt: 'Select the folder to generate disk' + selectedType + ' into',
-    value: os.homedir(),
+    value: location,
     ignoreFocusOut: true,
   });
   if (!selectedFolder) {
@@ -85,6 +87,9 @@ export async function buildDiskImage(image: any) {
     return;
   }
 
+  // store this path for later
+  await history.addImageBuild(image.name, selectedType, selectedFolder);
+
   return extensionApi.window.withProgress(
     { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + image.name },
     async progress => {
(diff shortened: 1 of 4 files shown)


deboer-tim:
Rebased on top of #70.

src/build-disk-image.ts
```diff
@@ -94,11 +99,11 @@ export async function buildDiskImage(image: any) {
 
       // Create log folder
       if (!fs.existsSync(selectedFolder)) {
-        await fs.mkdirSync(selectedFolder, { recursive: true });
+        fs.mkdirSync(selectedFolder, { recursive: true });
```
benoitf, line 102: might be better with `await fs.promises.mkdir`

src/history.ts
```diff
@@ -0,0 +1,92 @@
...
+        return;
+      }
+
+      const filePath = path.resolve(this.storagePath, filename);
+      if (!fs.existsSync(filePath)) {
+        return;
+      }
+
+      const infoBuffer = await readFile(filePath);
+      this.infos = JSON.parse(infoBuffer.toString('utf8'));
```
benoitf, line 51: and then you would received a string instead of a buffer and then no need to convert it using toString()

src/history.ts
```diff
@@ -0,0 +1,92 @@
...
+    } catch (err) {
+      console.error(err);
```
benoitf, line 53: don't know if it should be displayed back to the user like with an showErrorMessage ?

src/history.ts
```diff
@@ -0,0 +1,92 @@
...
+      this.infos.slice(0, 100);
+    }
+
+    this.saveFile().catch((err: unknown) => console.error('Unable to save history', err));
+  }
+
+  public async saveFile() {
+    try {
+      if (!fs.existsSync(this.storagePath)) {
+        await promisify(fs.mkdir)(this.storagePath);
```
benoitf, line 83: import from fs.promises rather than using promisify

src/build-disk-image.ts
deboer-tim: Switched in new commit.

src/history.ts
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { promisify } from 'node:util';
+
+const readFile = promisify(fs.readFile);
+const writeFile = promisify(fs.writeFile);
```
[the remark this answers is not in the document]
deboer-tim: Thanks. Switched in new commit, and made a note to update the PD code I copied from. :)

src/history.ts
```diff
@@ -0,0 +1,92 @@
...
+      if (!fs.existsSync(this.storagePath)) {
+        return;
+      }
+
+      const filePath = path.resolve(this.storagePath, filename);
+      if (!fs.existsSync(filePath)) {
+        return;
+      }
+
+      const infoBuffer = await readFile(filePath);
```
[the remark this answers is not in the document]
deboer-tim: Fixed in new commit.

src/history.ts
deboer-tim: Fixed in new commit.

src/history.ts
deboer-tim: Fixed in new commit.

src/history.ts
deboer-tim: For now the only user impact is 'we can't remember the path you used the last time you ran Podman Desktop', so I didn't feel like it's necessary to warn the user. If we use it for more critical functions then yes (although I'm also hoping there would be a move to OCI registry at some point).

cdrage (approved):
Was able to test this and code LGTM.

I do see we ignore some of the errors with history, but I'm okay with that for now.

https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/95614ed6-6694-444f-89e8-4525723a0033

Issue #48, opened 2024-01-17
Cleanup after successful build


deboer-tim:
If the build is successful, delete the container and volume.

If the build fails, we have the logs - but no great place to show the user (even if we can associate with the task, the full log is several hundred lines long). For now, we'll leave the exited container. (not sure if we should delete the volume)

PR #75, opened 2024-01-23
feat: clean containers and volumes up after a build

cdrage:
feat: clean containers and volumes up after a build

### What does this PR do?

* Removes the container after a success or unsuccessful build
* Removes any volumes created as a result of running the build

Other fixes:
* Moves the await show error message to outside of the try / catch as
  using 'await' on it was blocking the 'finally' statement regarding
  cleaning up
* Moves the "create container options" outside of the try / catch as it
  is not an async statement / it is just a generator

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/e2c6cdf1-cf29-412a-8e90-aae29a1a55e8

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/48

### How to test this PR?

1. Start a build
2. Stop the container mid-build
3. See the error output from building it
4. See that the container has been removed / any volumes removed / it's
   been cleaned up.

### DIFF
--- a/src/container-utils.ts
+++ b/src/container-utils.ts
@@ -101,18 +101,21 @@ export async function removeContainerIfExists(engineId: string, container: strin
   try {
     // List all the containers and check to see  if it exists
     const containers = await extensionApi.containerEngine.listContainers();
-    console.log('Containers: ', containers);
 
     // Find the one that matches the name we are looking for
     // The PD API is a bit weird in that it uses 'Names' for the container name instead of
     // Name and it's an array of strings with '/' appended to the beginning
     // So we need to check for that as well.
     let containerExists = false;
-    containers.forEach(c => {
-      if (c.Names.includes('/' + container)) {
-        containerExists = true;
-      }
-    });
+
+    // Only check for the container if there are containers
+    if (containers && containers.length > 0) {
+      containers.forEach(c => {
+        if (c.Names.includes('/' + container)) {
+          containerExists = true;
+        }
+      });
+    }
 
     // Delete the container if it exists
     if (containerExists) {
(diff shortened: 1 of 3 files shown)


src/container-utils.ts
```diff
@@ -123,3 +126,68 @@ export async function removeContainerIfExists(engineId: string, container: strin
...
   }
 }
+
+// Get all volume names that match the name of the container
+async function getVolumesMatchingContainer(engineId: string, container: string): Promise<string[]> {
+  try {
+    // Volumes returns a list of volumes across all engines. Only get the volume that matches our engineId
+    const engineVolumes = await extensionApi.containerEngine.listVolumes();
+    if (!engineVolumes) {
+      throw new Error('No volume providers found');
```
deboer-tim, line 136: Technically no volumes found? (could be providers)

src/container-utils.ts
```diff
@@ -123,3 +126,68 @@ export async function removeContainerIfExists(engineId: string, container: strin
...
+    }
+    // If none are found, just return an empty array / small warning
+    const volumes = engineVolumes.find(volume => volume.engineId === engineId);
+    if (!volumes) {
+      console.log('Ignoring removing volumes, no volume engine providers found for engineId: ', engineId);
```
deboer-tim, line 141: Likewise, this is just no volumes on the engine we care about.

src/container-utils.ts
```diff
@@ -123,3 +126,68 @@ export async function removeContainerIfExists(engineId: string, container: strin
...
+        });
+      });
+    });
+
+    // Remove any duplicates that may have been added (same volume used multiple times / mounted).
+    volumeNames = [...new Set(volumeNames)];
+    return volumeNames;
+  } catch (e) {
+    console.log(e);
+    throw new Error('There was an error getting the volumes: ' + e);
```
deboer-tim, line 164: If we have an error getting volumes this means we wouldn't even try to remove the container. IMHO we should still try, i.e. either don't throw this or put a catch around line 174.

cdrage:
@deboer-tim Updated based on the comments

src/container-utils.ts
cdrage: I've updated the function so that even if we error getting the volumes, we'll still try and remove the container.

deboer-tim (approved):

Issue #18, opened 2024-01-11
After creating disk image, and I choose not to launch VM, how do I re-launch?


cdrage:
If I create the disk image, output it to a directory, how do I now launch the VM using vfkit?

cdrage:
Closing this as we have added the "vfkit" button (although it's hidden right now..)

cdrage:
Closing this for now until we revisit vfkit

PR #83, opened 2024-01-24
feat: add launch action to bootable images

deboer-tim:
### What does this PR do?

Now that we clean up after a successful build, there's no way to launch vfkit on the image. This adds a command to bootable images, which tries to launch vfkit against the last disk image build.

### Screenshot / video of UI

Screenshot 2024-01-24 at 11 28 00 AM

Fixes part of #18.

### How to test this PR?

Launch the action on a bootable container that you haven't built before (it fails), then build an image and retry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,11 @@
           "command": "bootc.image.build",
           "title": "Build Disk Image",
           "when": "ostree.bootable in imageLabelKeys"
+        },
+        {
+          "command": "bootc.vfkit",
+          "title": "Launch as VM",
+          "when": "ostree.bootable in imageLabelKeys"
         }
       ],
       "dashboard/container": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -30,7 +30,7 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
 
   extensionContext.subscriptions.push(
     extensionApi.commands.registerCommand('bootc.vfkit', async container => {
-      await launchVFKit(container);
+      await launchVFKit(container, history);
     }),
 
     extensionApi.commands.registerCommand('bootc.image.build', async image => {
--- a/src/history.spec.ts
+++ b/src/history.spec.ts
@@ -58,3 +58,27 @@ test('check get returns latest after multiple adds', async () => {
 
   expect(history.getLastLocation()).toEqual('c2');
 });
+
+test('check lastBuild get updated', async () => {
+  vi.mock('node:fs', async () => {
+    return {
+      readFile: vi.fn().mockImplementation(() => '[]'),
+      writeFile: vi.fn().mockImplementation(() => Promise.resolve()),
+      existsSync: vi.fn().mockImplementation(() => true),
+    };
+  });
+
+  const history: History = new History('test');
+
+  await history.addImageBuild('a0', 'b0', 'c0');
+  await history.addImageBuild('a1', 'b1', 'c1');
+
+  expect(history.getLastBuildFor('a0').type).toEqual('b0');
+  expect(history.getLastBuildFor('a0').location).toEqual('c0');
+
+  await history.addImageBuild('a0', 'b0b', 'c0b');
+  await history.addImageBuild('a2', 'b2', 'c2');
+
+  expect(history.getLastBuildFor('a0').type).toEqual('b0b');
+  expect(history.getLastBuildFor('a0').location).toEqual('c0b');
+});
--- a/src/history.ts
+++ b/src/history.ts
@@ -54,9 +54,17 @@ export class History {
   public getLastLocation(): string | undefined {
     if (this.infos.length === 0) {
       return undefined;
-    } else {
-      return this.infos[0].location;
     }
+    return this.infos[0].location;
+  }
+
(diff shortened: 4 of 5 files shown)


cdrage (changes requested):
It worked!

Had no issues launching the previous image via vfkit. So all of this LGTM. Thanks for adding it to the image list instead.

Only change is we need to update the package.json as well to remove it from container page.

package.json
```diff
@@ -18,6 +18,11 @@
           "command": "bootc.image.build",
           "title": "Build Disk Image",
           "when": "ostree.bootable in imageLabelKeys"
+        },
+        {
+          "command": "bootc.vfkit",
```
cdrage, line 23: we should also remove the previous part about it being in container page: https://github.com/containers/podman-desktop-extension-bootc/blob/b275f8ff58183aaa2c2de0fea9821f1ea070623d/package.json#L28

cdrage (changes requested):

jwboyer:
I gave this PR a try because the build container is cleaned up too quickly so there's no opportunity to start it from that screen.  I merged it locally with #91 also included.  I built a disk image successfully and then tried to launch it and it failed.  This is the command it produced:

`main ↪️ [bootc-extension] --cpus,2,--memory,2048,--bootloader,efi,variable-store=./efi-variable-store,create,--device,virtio-blk,path=/Users/jboyer/images/jwboyer-eln,--device,virtio-serial,logFilePath=/Users/jboyer/images/vfkit-serial.log,--device,virtio-net,nat,mac=72:20:43:d4:38:62,--device,virtio-rng,--device,virtio-input,keyboard,--device,virtio-input,pointing,--device,virtio-gpu,width=1920,height=1080,--gui`

Note the virtio-blk,path specified didn't include the `...../images/disk.raw` part.

I looked at the code and it seems like we're passing the `selectedFolder` into history and not the actual image path.  I tried this diff and it at least tried to launch the VM:

```
diff --git a/src/build-disk-image.ts b/src/build-disk-image.ts
index 0854399..2a2e53b 100644
--- a/src/build-disk-image.ts
+++ b/src/build-disk-image.ts
@@ -88,7 +88,7 @@ export async function buildDiskImage(imageData: unknown, history: History) {
   }
 
   // store this path for later
-  await history.addImageBuild(image.name, selectedType, selectedFolder);
+  await history.addImageBuild(image.name, selectedType, imagePath);
 
   return extensionApi.window.withProgress(
     { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + image.name },
```

The VM doesn't actually boot, but at least the vfkit invocation is correct.

jwboyer:
FWIW, here's a screenshot of the VM after the changes mentioned above:

image

and the command that was run:
`main ↪️ [bootc-extension] --cpus,2,--memory,2048,--bootloader,efi,variable-store=./efi-variable-store,create,--device,virtio-blk,path=/Users/jboyer/images/jwboyer-eln/image/disk.raw,--device,virtio-serial,logFilePath=/Users/jboyer/images/jwboyer-eln/image/vfkit-serial.log,--device,virtio-net,nat,mac=72:20:43:d4:38:62,--device,virtio-rng,--device,virtio-input,keyboard,--device,virtio-input,pointing,--device,virtio-gpu,width=1920,height=1080,--gui`

Not sure if this is an issue with the disk itself, or something with vfkit

cdrage:
> FWIW, here's a screenshot of the VM after the changes mentioned above:
> 
> image
> and the command that was run: `main ↪️ [bootc-extension] --cpus,2,--memory,2048,--bootloader,efi,variable-store=./efi-variable-store,create,--device,virtio-blk,path=/Users/jboyer/images/jwboyer-eln/image/disk.raw,--device,virtio-serial,logFilePath=/Users/jboyer/images/jwboyer-eln/image/vfkit-serial.log,--device,virtio-net,nat,mac=72:20:43:d4:38:62,--device,virtio-rng,--device,virtio-input,keyboard,--device,virtio-input,pointing,--device,virtio-gpu,width=1920,height=1080,--gui`
> 
> Not sure if this is an issue with the disk itself, or something with vfkit

Press enter :)

This unfortunately is an issue with bootc-image-builder at the moment where it is getting "stuck" on the VM load: https://github.com/osbuild/bootc-image-builder/issues/105 and the screen isn't correctly refreshing.

They are at the moment working on improving the kernel parameters / templating the bootc image builder so that it would work correct.u

jwboyer:
> > Not sure if this is an issue with the disk itself, or something with vfkit
> 
> Press enter :)

I tried that before reporting 😄   Didn't do anything.

> This unfortunately is an issue with bootc-image-builder at the moment where it is getting "stuck" on the VM load: [osbuild/bootc-image-builder#105](https://github.com/osbuild/bootc-image-builder/issues/105) and the screen isn't correctly refreshing.
> 
> They are at the moment working on improving the kernel parameters / templating the bootc image builder so that it would work correctly.

OK.  I can be patient.  Thanks for the pointer.

deboer-tim:
As per comments elsewhere, we would love to provide options to test or run the disk image after building, but vfkit continued to have problems for users, and we have found no stable/open-source/compatible enough alternative. I'm finally withdrawing this PR, and we can follow up with new issues if there's another option.

PR #88, opened 2024-01-25
fix: change output folder

### LINKED ISSUE #86: Too easy to pick non-writable folder

When we prompt for the folder name it's easy to pick some random folder, only to see image builder fail because it couldn't write there.

See what we can do to make more folders writable for the mount, or at least error out early so the user isn't scratching their head looking at the build log.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Listing as a chore on our end, although I think we're confirming that the image builder really doesn't like writing to /tmp on some machines and that is causing builds to fail (but only when paired with certain output folders).

### Screenshot / video of UI

N/A

Fixes #86.

### How to test this PR?

Test any build.

### DIFF
--- a/src/build-disk-image.spec.ts
+++ b/src/build-disk-image.spec.ts
@@ -1,0 +1,48 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


cdrage (approved):
LGTM!

Pulled and tested, works.

So odd that /tmp doesn't work, but 🤷

Issue #97, opened 2024-01-26
add a release workflow


benoitf:
### Is your enhancement related to a problem? Please describe

releases should have their own Github Action workflow

### Describe the solution you'd like

add the release job to tag, build the image and publish it under the tag name and `latest`

deboer-tim:
v0.1.0 released, and everything appears to have worked - CI pushes tagged versions only to nightly, and release build does all the proper tagging (version + latest), pushing, and PRs version updating. The worst problem I can find is that the bot says "has been released." (missing what version was released :) will PR), so I'm counting this as done.

PR #107, opened 2024-01-29
feat: release build

deboer-tim:
### What does this PR do?

Adds a release build action in Github that should tag and create the release, create a PR to bump the version, build and publish the bits.

Copied almost verbatim from @benoitf's work on the Minikube extension.

### Screenshot / video of UI

N/A

See below, issue #97 will be fixed but I don't want to trigger immediate closure here.

### How to test this PR?

The normal way to test would be to copy the repo and test publishing to a private org, but this is fairly onerous, this is tested code from another repo, and failure isn't really going to affect anyone - so suggesting we merge, try it out, and rapidly fix if necessary.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,129 @@
+#
+# Copyright (C) 2023-2024 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,129 @@
...
+      - name: Login to ghcr.io
+        run: podman login --username ${{ github.repository_owner }} --password ${{ secrets.GITHUB_TOKEN }} ghcr.io
+
+      - name: Publish Image
+        id: publish-image
+        run: |
+          IMAGE_NAME=ghcr.io/${{ github.repository_owner }}/podman-desktop-extension-bootc
+          IMAGE_WITH_TAG=${IMAGE_NAME}:${{ steps.TAG_UTIL.outputs.bootcExtensionVersion }}
+          podman build -t $IMAGE_WITH_TAG .
+          podman push $IMAGE_WITH_TAG
```
benoitf, line 129: you should include `:latest` publication as well

benoitf (approved):

.github/workflows/release.yaml
deboer-tim: Updated.

PR #109, opened 2024-01-29
chore: remove vfkit from package.json

cdrage:
chore: remove vfkit from package.json

### What does this PR do?

Removes vfkit from package.json (for now) until we explore more stable
options / stability issues with it.

Right now we should be informing the user how to launch it via docs
instead.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Buttons should no longer be there.


deboer-tim (approved):
Beat me to it :).

PR #127, opened 2024-02-04
fix: change name

benoitf:
### What does this PR do?
bootc extension is published as `redhat.bootc` in the catalog so it should match the name of the package.json as well
or we change the catalog to use `redhat.bootc-extension`

so when installing the extension we're not finding the right extension after so some buttons are not replaced/ arestill in progress/etc

https://github.com/containers/podman-desktop/issues/5759

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "bootc-extension",
+  "name": "bootc",
   "displayName": "Bootable Container",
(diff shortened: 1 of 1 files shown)


benoitf:
if you could cut a new release after this change so we can update the catalog

deboer-tim (approved):
Will just check with Charlie and do a new release tomorrow.

PR #128, opened 2024-02-05
fix: pick unused container name

### LINKED ISSUE #103: Can't build two or more bootc drives at the same time
### Bug description

Currently it's not possible to successfully build two or more bootc drivers at the same time because the PD container that gets spun up to perform the action is always the same one, so when the second bootc drive starts it will overwrite the previous one that was running.

### Operating system

macOS

### Version

next (development version)

### Steps to reproduce

1. Start a bootc image build.
2. Wait until container starts running.
3. Start a second bootc image build.
4. Notice that no second container gets created, instead the original container gets recreated and as such destroying the initial build.
5. Notice that the task is left hanging in the PD task list.

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Make sure we pick an unused container name so that it's possible to kick off multiple builds at the same time.

Note that there are still two theoretical places this could still fail due to the delay between when you kick off the task and we check for conflicts vs when the container gets created: the first time you ever run (if the bootc builder image takes time to pull down) and if you're super-fast and run the action twice before the container engine can create a container and register back in Podman Desktop. The first seems very unlikely that someone would try, and the second I couldn't even reproduce when trying.

### Screenshot / video of UI

Screenshot 2024-02-05 at 8 25 41 AM

Fixes #103.

### How to test this PR?

Run two builds at the same time, confirm no name conflict and both images are produced.

### DIFF
--- a/src/build-disk-image.ts
+++ b/src/build-disk-image.ts
@@ -130,7 +130,7 @@ export async function buildDiskImage(imageData: unknown, history: History) {
       // Preliminary Step 0. Create the "bootc-image-builder" container
       // options that we will use to build the image. This will help with debugging
       // as well as making sure we delete the previous build, etc.
-      const buildImageContainer = createBuilderImageOptions(
+      const buildImageContainer = await createBuilderImageOptions(
         buildContainerName,
         image.name + ':' + image.tag,
         selectedType,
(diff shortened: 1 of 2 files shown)


src/build-disk-image.ts
```diff
@@ -233,17 +233,39 @@ async function logContainer(image, containerId: string, progress, callback: (dat
...
+  let count = 2;
+  while (containers.includes(unusedName)) {
+    unusedName = name + '-' + count++;
+  }
+  return unusedName;
+}
+
 // Create builder options for the "bootc-image-builder" container
-export function createBuilderImageOptions(
+export async function createBuilderImageOptions(
```
cdrage, line 259: I would prefer if we keep this non-async as all it does is generate the create options / not touching this function within this PR.

Instead, we should have getUnusedName outside of createBuilderImageOptions to get the name, and then simply pass it into createBuilderImageOptions.

src/build-disk-image.ts
```diff
@@ -130,7 +130,7 @@ export async function buildDiskImage(imageData: unknown, history: History) {
       // Preliminary Step 0. Create the "bootc-image-builder" container
       // options that we will use to build the image. This will help with debugging
       // as well as making sure we delete the previous build, etc.
-      const buildImageContainer = createBuilderImageOptions(
+      const buildImageContainer = await createBuilderImageOptions(
```
cdrage, line 133: IMO we should do something like:

```ts
const containerName = await getUnusedName(buildContainerName)
```

Then pass it into createBuilderImageOptions

cdrage (changes requested):

src/build-disk-image.ts
deboer-tim: Yeah, that's better. Fixed in new commit.

src/build-disk-image.ts
deboer-tim: Yeah, that's better. Fixed in new commit.

cdrage (approved):

PR #144, opened 2024-02-13
feat: remove old builder images

### LINKED ISSUE #98: Clean up old builder images

As image builder has new releases and we move up to them (e.g. #89), this automatically pulls the new image and uses it - but never cleans up the old one.

It's not ideal, but I think before pulling we should check if there are any other versions of the image builder and delete them.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds a new function to delete images with a given name that do not have the same tag (nor a second tag; you can tag to keep something). Uses this to look for and clean up any old builder images, so that we'll automatically clean up whenever we pull a new one.

### Screenshot / video of UI

N/A

Fixes #98.

### How to test this PR?

Pull the 'latest' bootc builder image and then build a new disk image, 'latest' will be deleted since we're using a tagged one.

### DIFF
--- a/src/container-utils.ts
+++ b/src/container-utils.ts
@@ -56,6 +56,60 @@ export async function pullImage(image: string) {
   }
 }
 
+// Delete all copies of the given image except for the current one
+export async function deleteOldImages(engineId: string, currentImage: string) {
+  console.log('Deleting old images: ', currentImage);
+  try {
+    // List all the images and check to see if it exists
+    const images = await extensionApi.containerEngine.listImages();
+    if (!images || images.length === 0) {
+      return;
+    }
+
+    // We're looking to delete images that have the same name but different tags
+    const indexTag = currentImage.lastIndexOf(':');
+    const currentName = currentImage.slice(0, indexTag);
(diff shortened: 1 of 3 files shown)


src/container-utils.ts
```diff
@@ -56,6 +56,60 @@ export async function pullImage(image: string) {
...
+    });
+
+    // Delete the images
+    await imageIdsToRemove.reduce((prev: Promise<void>, imageId) => {
+      return prev
+        .then(() => extensionApi.containerEngine.deleteImage(engineId, imageId))
+        .catch((e: unknown) => console.error('error while removing image', e));
+    }, Promise.resolve());
+  } catch (e) {
+    console.error(e);
```
cdrage, line 108: we throw so no need for console error?

cdrage (approved):
One small change, otherwise LGTM! Worked well and it successfully deleted the previous copies.

src/container-utils.ts
deboer-tim: Actually, if this goes wrong we don't want the rest to fail, so I've removed the throw instead.

Issue #167, opened 2024-03-04
Check for rootful mode changed in Podman 5.0


cdrage:
### Bug description

In Podman 5.0, the configuration changed and no longer has a "Host" subkey:

```json
▶ podman machine inspect
[
     {
          "ConfigDir": {
               "Path": "/Users/cdrage/.config/containers/podman/machine/applehv"
          },
          "ConnectionInfo": {
               "PodmanSocket": {
                    "Path": "/Users/cdrage/.local/share/containers/podman/machine/applehv/podman.sock"
               },
               "PodmanPipe": null
          },
          "Created": "2024-03-01T13:48:37.611598-05:00",
          "LastUp": "2024-03-01T13:56:15.665784-05:00",
          "Name": "podman-machine-default",
          "Resources": {
               "CPUs": 5,
               "DiskSize": 100,
               "Memory": 2048,
               "USBs": []
          },
          "SSHConfig": {
               "IdentityPath": "/Users/cdrage/.local/share/containers/podman/machine/machine",
               "Port": 59284,
               "RemoteUsername": "core"
          },
          "State": "running",
          "UserModeNetworking": true,
          "Rootful": true
     }
]
```

Instead "Rootful": true is in a different area.

### Operating system

mac m1 and podman 5.0-dev

### Version

next (development version)

### Steps to reproduce

Go to try and build a disk image.

It'll fail as it'll check for rootful mode but then fail because the configuration json has changed.

PR #172, opened 2024-03-05
chore: fix rootful mode for podman 5.0.0 / enables linux check

cdrage:
chore: fix rootful mode for podman 5.0.0 / enables linux check

### What does this PR do?

* Enables being able to test against podman 5.0.0
* Re-enables the Linux check now that you can create a podman machine on
  Linux

### Screenshot / video of UI

N/A. You should be able to build like normal.

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/167

### How to test this PR?

1. Test against Podman 5.0.0
2. Be able to build an image like usual.

### DIFF
--- a/packages/backend/src/machine-utils.spec.ts
+++ b/packages/backend/src/machine-utils.spec.ts
@@ -84,3 +84,77 @@ test('Check isPodmanMachineRootful functionality', async () => {
   spyReadFile.mockResolvedValue(JSON.stringify({ Rootful: true }));
   await expect(machineUtils.isPodmanMachineRootful()).resolves.toBe(true);
 });
+
+test('Fail isPodmanMachineRootful functionality be false if Rootful does not exist in the root of the object, or in HostUser', async () => {
+  const fakeMachineInfoJSON = {
+    Host: {
+      Arch: 'amd64',
+      CurrentMachine: '',
+      DefaultMachine: '',
+      EventsDir: 'dir1',
+      MachineConfigDir: 'dir2',
+      MachineImageDir: 'dir3',
+      MachineState: '',
+      NumberOfMachines: 5,
+      OS: 'windows',
+      VMType: 'wsl',
+    },
+  };
+
+  vi.spyOn(extensionApi.process, 'exec').mockImplementation(
+    () =>
+      new Promise<extensionApi.RunResult>(resolve => {
+        resolve({ stdout: JSON.stringify(fakeMachineInfoJSON) } as extensionApi.RunResult);
+      }),
+  );
+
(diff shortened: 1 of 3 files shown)


packages/backend/src/machine-utils.spec.ts
```diff
@@ -84,3 +84,77 @@ test('Check isPodmanMachineRootful functionality', async () => {
...
+  vi.mock('node:fs');
+  vi.spyOn(fs, 'existsSync').mockImplementation(() => {
+    return true;
+  });
+
+  // Mock the readFile function to return the "fake" file with Rootful not existing
+  const spyReadFile = vi.spyOn(fs.promises, 'readFile');
+
+  // Mock reading the file to have Rootful as true
+  spyReadFile.mockResolvedValue(JSON.stringify({ HostUser: { Rootful: true } }));
```
deboer-tim, line 158: Why not put rootful in the fake info above? (Or can we mock readMachineConfig() directly to just return the JSON here?)

deboer-tim (approved):
It seems like the test could be simpler, but code LGTM.

FWIW Brent said the machine config file is versioned starting with 5.0 (i.e. no version = Podman <5.0, v 1.0 = Podman 5.0), but looking for both config properties is simpler and one less thing to go wrong. 👍🏼

packages/backend/src/machine-utils.spec.ts
cdrage: See below:

```json
{
  "Host": {
    "Arch": "arm64",
    "CurrentMachine": "podman-machine-default",
    "DefaultMachine": "",
    "EventsDir": "/var/folders/cn/pm5mgrcs5h339nphypxtvd080000gn/T/storage-run-501/podman",
    "MachineConfigDir": "/Users/cdrage/.config/containers/podman/machine/applehv",
    "MachineImageDir": "/Users/cdrage/.local/share/containers/podman/machine/applehv",
    "MachineState": "Running",
    "NumberOfMachines": 1,
    "OS": "darwin",
    "VMType": "applehv"
  },
  "Version": {
    "APIVersion": "5.0.0-dev",
    "Version": "5.0.0-dev",
    "GoVersion": "go1.21.6",
    "GitCommit": "0ace66f1aed453ebc25febbfc79636095d08ce78",
    "BuiltTime": "Fri Mar  1 13:46:53 2024",
    "Built": 1709318813,
    "OsArch": "darwin/arm64",
    "Os": "darwin"
  }
}
```

For podman machine, we use `podman machine info --format json` which gets the directory of where the actual json information is. 

That's why we fake it as the isPodmanRootful command first gets that.

PR #175, opened 2024-03-05
feat: refactor of build logic / more history info / quickpicks

### LINKED ISSUE #173: Split out "quickpick" login into separate file so that webview can use functions.

Right now in: https://github.com/containers/podman-desktop-extension-bootc/blob/main/packages/backend/src/build-disk-image.ts#L31

We have all the "quickpick" login in the same function.

In order to use the same function / other functions such as a webview, this section needs to be refactored.

Have quick-pick selection in a separate file

_No response_

### Additional context

_No response_

### LINKED ISSUE #174: History should save more information such as timestamp, engineId, build container ID, etc.

When saving the history, we should include a lot more information such as the build container ID, the timestamp, engineID, etc.

Save information similar to this:

```
export interface BootcBuildInfo {
  name: string;
  tag: string;
  engineId: string;
  type: string;
  folder: string;
  arch: string;
  status?: BootcBuildStatus;
  timestamp?: string;
  buildContainerId?: string; // The image ID that is used to build the image
}
```

_No response_

### Additional context

_No response_

cdrage:
feat: refactor of build logic / more history info / quickpicks

### What does this PR do?

In order to add the "quickpick" logic in a separate file, it was
beneficial to also refactor what history information we also save.

In this PR we:
* Separate the "quickpick" functionality into a separate file, so we may
  use the build disk image function separately
* Additional history information being saved such as build container id,
  timestamp, etc.

### Screenshot / video of UI

N/A, internal only. Build functionality should work as intended.

Closes #173
Closes #174

### How to test this PR?

Build a disk image, it should continue as normal.

Check:
`~/.local/share/containers/podman-desktop/extensions-storage/redhat.bootc/history.json`
for the 'updated history info'

### DIFF
--- a/packages/backend/src/build-disk-image.ts
+++ b/packages/backend/src/build-disk-image.ts
@@ -18,123 +18,100 @@
 
 import type { ContainerCreateOptions } from '@podman-desktop/api';
 import * as extensionApi from '@podman-desktop/api';
-import * as os from 'node:os';
 import * as fs from 'node:fs';
 import { resolve } from 'node:path';
 import * as containerUtils from './container-utils';
-import * as machineUtils from './machine-utils';
 import { bootcImageBuilderContainerName, bootcImageBuilderName } from './constants';
+import type { BootcBuildInfo } from '@shared/src/models/bootc';
 import type { History } from './history';
+import * as machineUtils from './machine-utils';
 
 const telemetryLogger = extensionApi.env.createTelemetryLogger();
 
-export async function buildDiskImage(imageData: unknown, history: History) {
-  // Before we do ANYTHING, we should be checking to see if the podman machine is rootful or not
-  // as that's a requirement for bootc-image-builder to work correctly.
-  const isRootful = await machineUtils.isPodmanMachineRootful();
-  if (!isRootful) {
-    await extensionApi.window.showErrorMessage(
-      'The podman machine is not set as rootful. Please recreate the podman machine with rootful privileges set and try again.',
-    );
-    return;
-  }
-
-  const image = imageData as { name: string; engineId: string; tag: string };
+export async function buildDiskImage(build: BootcBuildInfo, history: History): Promise<void> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const telemetryData: Record<string, any> = {};
+  let errorMessage: string;
 
-  const selection = await extensionApi.window.showQuickPick(
-    [
-      { label: 'QCOW2', detail: 'QEMU image (.qcow2)', format: 'qcow2' },
-      { label: 'AMI', detail: 'Amazon Machine Image (.ami)', format: 'ami' },
-      { label: 'RAW', detail: 'Raw image (.raw) with an MBR or GPT partition table', format: 'raw' },
-      { label: 'ISO', detail: 'ISO standard disk image (.iso) for flashing media and using EFI', format: 'iso' },
-    ],
-    {
-      title: 'Select the type of disk image to create',
-    },
-  );
-  if (!selection) {
-    telemetryData.canceled = true;
-    telemetryLogger.logUsage('buildDiskImage', telemetryData);
-    return;
-  }
-  const selectedType = selection.format;
-  telemetryData.imageType = selectedType;
+  const requiredFields = [
+    { field: 'name', message: 'Bootc image name is required.' },
+    { field: 'tag', message: 'Bootc image tag is required.' },
+    { field: 'type', message: 'Bootc image type is required.' },
+    { field: 'engineId', message: 'Bootc image engineId is required.' },
+    { field: 'folder', message: 'Bootc image folder is required.' },
+    { field: 'arch', message: 'Bootc image architecture is required.' },
+  ];
 
-  const selectionArch = await extensionApi.window.showQuickPick(
-    [
-      { label: 'ARM64', detail: 'ARM® aarch64 systems', arch: 'arm64' },
-      { label: 'AMD64', detail: 'Intel and AMD x86_64 systems', arch: 'amd64' },
-    ],
-    {
-      title: 'Select the architecture',
-    },
-  );
-  if (!selectionArch) {
-    telemetryData.canceled = true;
-    telemetryLogger.logUsage('buildDiskImage', telemetryData);
-    return;
+  // VALIDATION CHECKS
+  for (const { field, message } of requiredFields) {
+    if (!build[field]) {
+      await extensionApi.window.showErrorMessage(message);
+      throw new Error(message);
+    }
   }
-  const selectedArch = selectionArch.arch;
-  telemetryData.arch = selectedArch;
 
-  const location = history.getLastLocation() || os.homedir();
-  const selectedFolder = await extensionApi.window.showInputBox({
-    prompt: 'Select the folder to generate disk' + selectedType + ' into',
-    value: location,
-    ignoreFocusOut: true,
-  });
-  if (!selectedFolder) {
-    telemetryData.canceled = true;
-    telemetryLogger.logUsage('buildDiskImage', telemetryData);
-    return;
+  const isRootful = await machineUtils.isPodmanMachineRootful();
+  if (!isRootful) {
+    const errorMessage =
+      'The podman machine is not set as rootful. Please recreate the podman machine with rootful privileges set and try again.';
+    await extensionApi.window.showErrorMessage(errorMessage);
+    throw new Error('The podman machine is not set as rootful.');
   }
-  // Make this into a map that 'qcow2' -> 'disk.qcow2'
-  // and 'ami' -> 'qcow2/disk.raw'
-  // and 'raw' -> 'image/disk.raw'
-  // and 'iso' -> 'bootiso/disk.iso'
-  // and then use that to build the path
-  const imageNameMap = {
-    qcow2: 'qcow2/disk.qcow2',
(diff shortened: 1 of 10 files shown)


cdrage:
FYI rebased against https://github.com/containers/podman-desktop-extension-bootc/pull/172 and requires that PR to be merged in first.

packages/shared/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "allowJs": true,
+    "checkJs": true,
+  },
+  "include": [
+    "src/**/*.d.ts",
+    "src/**/*.ts",
+    "src/**/*.js",
+    "types/*.d.ts",
+    "../../types/**/*.d.ts",
+  ]
```
benoitf, line 25: shared should not have any svelte imports (basically shared can't have dom or node types it's shared between server side and client side)

packages/shared/vitest.config.js
```diff
@@ -0,0 +1,36 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (C) 2024 Red Hat, Inc.
```

packages/backend/src/quickpicks.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const imageNameMap = {
+    qcow2: 'qcow2/disk.qcow2',
+    ami: 'image/disk.raw',
+    raw: 'image/disk.raw',
+    iso: 'bootiso/disk.iso',
+  };
+
+  console.log('selectedType', selectedType);
+  console.log('selectedFolder', selectedFolder);
```
benoitf, line 65: remove ?

packages/backend/src/quickpicks.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+  );
+  if (!selectionArch) {
+    return;
+  }
+  const selectedArch = selectionArch.arch;
+
+  const location = history.getLastFolder() || os.homedir();
+  const selectedFolder = await extensionApi.window.showInputBox({
+    prompt: 'Select the folder to generate disk ' + selectedType + ' into',
```
benoitf, line 48:
```suggestion
    prompt: `Select the folder to generate disk ${selectedType} into`,
```

packages/backend/src/quickpicks.ts
```diff
@@ -0,0 +1,74 @@
...
+    {
+      title: 'Select the architecture',
+    },
+  );
+  if (!selectionArch) {
+    return;
+  }
+  const selectedArch = selectionArch.arch;
+
+  const location = history.getLastFolder() || os.homedir();
```
benoitf, line 46:
```suggestion
  const location = history.getLastFolder() ?? os.homedir();
```

packages/backend/src/quickpicks.ts
```diff
@@ -0,0 +1,74 @@
...
+  folder: string;
+  path: string;
+  arch: string;
+}
+
+export async function bootcBuildOptionSelection(history: History): Promise<BootcBuildOptionSelection | undefined> {
+  const selection = await extensionApi.window.showQuickPick(
+    [
+      { label: 'QCOW2', detail: 'QEMU image (.qcow2)', format: 'qcow2' },
+      { label: 'AMI', detail: 'Amazon Machine Image (.ami)', format: 'ami' },
```
benoitf, line 17: I think it might work better having a Class with history being a class field

and then you call the method to get the build options

packages/backend/src/history.ts
```diff
@@ -51,39 +46,54 @@ export class History {
...
+    } catch (err) {
+      console.error('Error saving file:', err);
+    }
+  }
+
+  public getLastFolder(): string | undefined {
+    if (this.infos.length === 0) {
+      return undefined;
+    } else {
+      return this.infos[0].folder;
```
benoitf, line 96: would it work with just `this.infos?.[0]?.folder`

packages/backend/src/history.ts
```diff
@@ -51,39 +46,54 @@ export class History {
...
+          info.type === buildInfo.type &&
+          info.arch === buildInfo.arch &&
+          info.tag === buildInfo.tag
+        ),
+    );
+    await this.saveFile();
   }
 
-  public async saveFile() {
+  private async saveFile(): Promise<void> {
```
benoitf, line 82: maybe use protected to better test it ?

packages/backend/src/history.ts
```diff
@@ -18,21 +18,16 @@
...
 
-interface ImageInfo {
-  image: string;
-  type: string;
-  location: string;
-}
-
 export class History {
-  infos: ImageInfo[] = [];
+  public infos: BootcBuildInfo[] = [];
```
benoitf, line 26: should not be public

packages/backend/src/build-disk-image.ts
```diff
@@ -211,33 +217,68 @@ export async function buildDiskImage(imageData: unknown, history: History) {
...
         );
       } else {
+        try {
+          // Update the image build status to error
+          build.status = 'error';
+          await history.addOrUpdateBuildInfo(build);
+        } catch (e) {
+          // Same as above, do not want to block other parts of the build
+          // so just output to console.
+          console.error('Error updating image build status to error', e);
```
benoitf, line 254: I think it would be interesting to know what build it is in the error

else from end user POV it's not really helpful when you read 'Error updating image build'

packages/backend/src/build-disk-image.ts
```diff
@@ -150,51 +122,85 @@ export async function buildDiskImage(imageData: unknown, history: History) {
...
+          }
+        }
 
         // If we get here, the container has exited with a zero exit code
         // it's successful as well so we will write the log file
         successful = true;
         telemetryData.success = true;
-      } catch (error) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (error: any) {
```
benoitf, line 203: should be unknown for errors

packages/backend/src/build-disk-image.ts
```diff
@@ -150,51 +122,85 @@ export async function buildDiskImage(imageData: unknown, history: History) {
...
+          await containerUtils.waitForContainerToExit(containerId);
+        } catch (error) {
+          // If we error out, BUT the container does not exist in the history, we will silently error
+          // as it's possible that the container was removed by the user during the build cycle / deleted from history.
+
+          // Check if history.infos has an entry with a containerId
+          const historyExists = history.infos.some(info => info.buildContainerId === containerId);
+          if (!historyExists) {
+            console.error(
+              'Container has errored out, but there is no history entry for it. This is likely due to the container being removed intentionally during the build cycle. Ignore this. Error:',
```
benoitf, line 189: IMHO from end user, you can't really say 'for it'
I would expect to have message about for example the image being built or the container name, etc

packages/backend/src/build-disk-image.ts
```diff
@@ -150,51 +122,85 @@ export async function buildDiskImage(imageData: unknown, history: History) {
       const containerName = await getUnusedName(buildContainerName);
       const buildImageContainer = createBuilderImageOptions(
         containerName,
-        image.name + ':' + image.tag,
-        selectedType,
-        selectedArch,
-        selectedFolder,
+        build.name + ':' + build.tag,
```
benoitf, line 125:
```suggestion
        `${build.name}:${build.tag}`,
```

packages/backend/src/build-disk-image.ts
```diff
@@ -18,128 +18,100 @@
...
+      logData += 'Image:  ' + build.name + '\n';
+      logData += 'Type:   ' + build.type + '\n';
+      logData += 'Folder: ' + build.folder + '\n';
       logData += '----------\n';
 
       // Create log folder
-      if (!fs.existsSync(selectedFolder)) {
-        fs.mkdirSync(selectedFolder, { recursive: true });
+      if (!fs.existsSync(build.folder)) {
+        fs.mkdirSync(build.folder, { recursive: true });
```
benoitf, line 112: await fs.promises.mkdir

packages/shared/tsconfig.json
cdrage: I tried modifying this but gave up on the erroring out, I implemented based on your PR a while back: https://github.com/projectatomic/ai-studio/commit/0ea4d2135f4e15402d4bfc07ddc3e43c8fa136a1#diff-28fbbd10733902283f23978b34aacb8284bc7af50d729f6d0ea1526ba625d322

Any suggestions how we could modify this / still have it pass / make the changes too to aistudio?

packages/backend/src/quickpicks.ts
cdrage: I thought of that too, but then I realized that we'd have to pass in the extensionContext in order to get the storagePath information, so it'd be trading one thing for another.

cdrage:
Thanks for the thorough review! I've updated the PR @benoitf

cdrage:
I'm going to merge this since we've gotten one review and all functionalities are still working. This will help in preparation for the "build view" to be integrated.

packages/shared/tsconfig.json
benoitf: I still don't see why shared has svelte in its tsconfig file

cdrage:
@benoitf you're right, after I cleared cache and removed the svelte information from tsconfig, it worked now 👍

benoitf (approved):

Issue #150, opened 2024-02-16
Welcome page (initial webview)


deboer-tim:
### Is your enhancement related to a problem? Please describe

We're switching to use a webview instead of a quickpick for selecting the options to start a build. If we're using a webview anyway, we thought it would make a good place to start to explore bootable images and how to find the action.

### Describe the solution you'd like

Create a welcome/homepage webview for bootable containers. To start, it would provide an intro to bootable containers, tell the user how to get started, and link to documentation.

Sample for general design:
<img width="747" alt="Screenshot 2024-02-15 at 1 29 46 PM" src="https://github.com/containers/podman-desktop-extension-bootc/assets/19958075/d49e8b17-59ef-483c-9958-7612546b95ac">

cdrage:
Updated the issue to reflect it'll be the "dashboard" as well and assigning it to myself as I've got some POC written

cdrage:
Splitting this issue into two. This issue covers the "Welcome" section, issue #196 covers dashboard.

cdrage:
@deboer-tim @slemeur 

#197 is in-review and feature complete.

We do not have UX/UI mockups, but it was talked about in previous calls that we would have the middle part of something like this:

![Screenshot 2024-03-20 at 7 40 10 PM](https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/850022e4-214a-48bc-94ab-efe656781b4b)

Should we have it in the center when there are no builds completed?

![Screenshot 2024-03-20 at 7 40 50 PM](https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/51cadfb9-67b9-47a0-8a3c-ac744fcae5c3)

This would make sense as if the user has done absolutely zero builds, the generic "no builds shown" should point to our README (the github page).

Let me know your thoughts.

nichjones1:
Need a small mockup for “Welcome Page” for bootc that incorporates your “dashboard” for past X builds

ekidneyrh:
I reused the Learning Center cards. Here's what that would look like: 

No builds:
![Screenshot from 2024-03-22 10-49-15](https://github.com/containers/podman-desktop-extension-bootc/assets/87311656/b7813c10-8241-47db-9d41-4dfb965f2c1b)

Builds with links visible:
[...]

PR #192, opened 2024-03-14
feat: add store for history

### LINKED ISSUE #188: Add history stores / subscriber, to use within the dashboard / welcome screen.
### Is your feature request related to a problem? Please describe

Smaller issue related to #150 

To be able to view the past history, we should add a store / subscriber so we can see what has changed / etc. 

_No response_

### Additional context

_No response_

cdrage:
feat: add store for history

### What does this PR do?

Adds a notification system / store for history so the frontend can be
notified and correctly retrieve the file history.

It "refreshes" based on file change, deletion or creation.

### Screenshot / video of UI

N/A

Closes #188

Part of #150

### How to test this PR?

- [x] Covered by tests


packages/shared/src/models/bootc.ts
```diff
@@ -29,3 +29,5 @@ export interface BootcBuildInfo {
 }
 
 export type BootcBuildStatus = 'running' | 'creating' | 'success' | 'error' | 'lost' | 'deleting';
+
+export const BootcHistoryFilename = 'history.json';
```
benoitf, line 33: I think variables or const should be either all UPPER_CASE or just camelCase here it's more like a Class/Interface/Type

packages/frontend/src/stores/rpcReadable.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  function debounce(func: () => Promise<T>): () => Promise<T> {
+    return () =>
+      new Promise<T>(resolve => {
+        if (timeoutId) {
+          clearTimeout(timeoutId);
+          timeoutId = undefined;
+        }
+
+        // throttle timeout, ask after 5s to update anyway to have at least UI being refreshed every 5s if there is a lot of events
```
deboer-tim, line 68: I really doubt this would be hit often... but if it is, 5s is a long time for no UI updates. Maybe cut this down to like 2s?

packages/shared/src/models/bootc.ts
```diff
@@ -29,3 +29,5 @@ export interface BootcBuildInfo {
 }
 
 export type BootcBuildStatus = 'running' | 'creating' | 'success' | 'error' | 'lost' | 'deleting';
+
+export const BOOTC_HISTORY_FILENAME = 'history.json';
```
deboer-tim, line 33: Why is this in /shared? I don't see it used in frontend, and seems like something that should be kept within backend (or if it was shared, then frontend could poll the file itself...).

packages/frontend/src/stores/history.ts
```diff
@@ -0,0 +1,11 @@
+import type { Readable } from 'svelte/store';
+import { Messages } from '/@shared/src/messages/Messages';
+import { bootcClient } from '/@/api/client';
+import type { BootcBuildInfo } from '/@shared/src/models/bootc';
+import { RPCReadable } from '/@/stores/rpcReadable';
+
+export const historyInfo: Readable<BootcBuildInfo[]> = RPCReadable<BootcBuildInfo[]>(
```
deboer-tim, line 7: Could you rename the file to historyInfo.ts? That would both match the const and filename, and avoid accidentally importing or talking about the 'wrong' history.ts.

packages/backend/src/history/historyNotifier.ts
```diff
@@ -0,0 +1,48 @@
...
+    private readonly storagePath: string,
+  ) {
+    this.#watcher = podmanDesktopApi.fs.createFileSystemWatcher(path.join(this.storagePath, BOOTC_HISTORY_FILENAME));
+    this.#watcher.onDidChange(this.notify.bind(this));
+    this.#watcher.onDidCreate(this.notify.bind(this));
+    this.#watcher.onDidDelete(this.notify.bind(this));
+  }
+
+  async notify(): Promise<void> {
+    await this.webview.postMessage({
```
deboer-tim, line 39: This creates an odd circular dependency where a change to the file keeps the cached data in backend but triggers an event to the frontend, which calls back to listHistoryInfo() to get the info, which triggers the actual file reload. The file is reloaded on any call from the frontend, even if it hasn't changed.

I know this is unlikely to cause issues in reality, but it seems like this function should be the one triggering a reload (which wouldn't be great b/c it could reload lots even if we're not using the history atm), or better yet just clearing the history and having history.getHistory() automatically reload when there's no data. i.e. history.getHistory() is a little smarter and the current forced reload in listHistoryInfo() could be removed.

deboer-tim (changes requested):
Good function t have, but raised a couple points on dependencies and a question.

packages/shared/src/models/bootc.ts
cdrage: Used in both history.ts and historyNotifier.ts so shared was going to be the best place for it, as well as if we had to do anything in the frontend.

packages/frontend/src/stores/rpcReadable.ts
cdrage: copied over from aistudio, I'd like to keep it to 5s as it's a direct copy / we should honestly have a central area to keep track of shared code.

packages/backend/src/history/historyNotifier.ts
cdrage: Needed in order to get RPCReadable working between frontend and backend. The notifier allows the frontend to check / pull the newest one, rather than using the backend. I tried implementing it the opposite way (like you had mentioned).. but ran into the opposite which was caching issues.

This is similar to the aistudio implementation. and is intentional to not post in values and used to just notify the frontend, we end up retrieving the file with listHistoryInfo like what you said in here:

```
export const historyInfo: Readable<BootcBuildInfo[]> = RPCReadable<BootcBuildInfo[]>(
  [],
  [Messages.MSG_HISTORY_UPDATE],
  bootcClient.listHistoryInfo,
);
```

With stores, the only time that it is actually reloaded is the initial loading (when switching to webview) or the notify from postMessage.

cdrage:
Updated with changes!

packages/frontend/src/stores/rpcReadable.ts
deboer-tim: Ok, better to keep equivalent code for now.

packages/shared/src/models/bootc.ts
deboer-tim: I would prefer it stayed in backend - odd to tell the frontend where a file is, but then expect it to use backend API and notifier. Better that it knows nothing about where/how the history is saved.

packages/frontend/src/stores/history.ts
deboer-tim: Done in new commit.

packages/backend/src/history/historyNotifier.ts
deboer-tim: I'm not suggesting posting values, just that I don't like the pattern of telling the frontend that there is a change but keeping cached values until frontend asks you for data (and forced reload either way). One path watching for file system changes and forcing a reload either way, but others like quickpick using 'cached' data.

I am fine approving as-is since I can't find any path that would fail today except if you touch the file directly, but I think we need to track the inconsistency since it could lead to odd failures in the future if you're not aware of it.

cdrage:
@deboer-tim Moved the const to `history.ts` instead of the shared folder, ready for a re-review

deboer-tim (approved):
Thanks for the changes, LGTM.

packages/shared/src/models/bootc.ts
deboer-tim: Fixed.

PR #205, opened 2024-03-20
chore: dynamically write log files

### LINKED ISSUE #78: Generate image-build.log dynamically

We don't generated the image build log until the end.

It would be useful to have it stream even periodically to disk.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Write the log files to disk as we get the data, instead of buffering and waiting until the end. This is better for debugging outside of Podman Desktop and will help in the future if we need to show logs in another view.

### Screenshot / video of UI

N/A

Fixes #78.

### How to test this PR?

Build an image and open image-build.log before it is done.

### DIFF
--- a/packages/backend/src/build-disk-image.ts
+++ b/packages/backend/src/build-disk-image.ts
@@ -130,6 +130,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
       );
       logData += JSON.stringify(buildImageContainer, undefined, 2);
       logData += '\n----------\n';
+      try {
+        fs.writeFileSync(logPath, logData);
+      } catch (e) {
+        // ignore
+      }
 
       if (!buildImageContainer) {
         await extensionApi.window.showErrorMessage('Error creating container options.');
@@ -168,7 +173,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
 
         // Step 3.1 Since we have started the container, we can now go get the logs
         await logContainer(build.engineId, containerId, progress, data => {
-          logData += data;
+          try {
+            fs.appendFileSync(logPath, data);
+          } catch (e) {
+            // ignore
+          }
         });
 
         // Step 4. Wait for the container to exit
@@ -204,13 +213,6 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
         console.error(error);
         telemetryData.error = error;
       } finally {
-        // Regardless, write the log file and ignore if we can't even write it.
-        try {
-          fs.writeFileSync(logPath, logData, { flag: 'w' });
-        } catch (e) {
-          // ignore
-        }
-
         // ###########
         // # CLEANUP #
         // ###########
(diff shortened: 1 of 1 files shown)


packages/backend/src/build-disk-image.ts
```diff
@@ -130,6 +130,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
       );
       logData += JSON.stringify(buildImageContainer, undefined, 2);
       logData += '\n----------\n';
+      try {
+        fs.writeFileSync(logPath, logData);
```
benoitf, line 134: maybe use the promisify version from 'node:fs/promises'. fs.writeFile(

packages/backend/src/build-disk-image.ts
```diff
@@ -168,7 +173,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
 
         // Step 3.1 Since we have started the container, we can now go get the logs
         await logContainer(build.engineId, containerId, progress, data => {
-          logData += data;
+          try {
+            fs.appendFileSync(logPath, data);
```
benoitf, line 177: node:fs/promises appendFile

packages/backend/src/build-disk-image.ts
```diff
@@ -168,7 +173,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
...
+          } catch (e) {
+            // ignore
```
benoitf, line 179: would be nice to log as console.debug in case you need to know why it's failing

packages/backend/src/build-disk-image.ts
```diff
@@ -130,6 +130,11 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
       );
       logData += JSON.stringify(buildImageContainer, undefined, 2);
       logData += '\n----------\n';
+      try {
+        fs.writeFileSync(logPath, logData);
+      } catch (e) {
+        // ignore
```
benoitf, line 136: would be nice to log as console.debug in case you need to know why it's failing

packages/backend/src/build-disk-image.ts
deboer-tim: Switched to fs.promises.writeFile(), since we're using fs.promises already (could have both imports but can't remove regular import due to use of existsSync()).

packages/backend/src/build-disk-image.ts
deboer-tim: Switched.

packages/backend/src/build-disk-image.ts
deboer-tim: Fixed.

packages/backend/src/build-disk-image.ts
deboer-tim: Fixed.

benoitf (approved):
FYI I don't see the buildDiskImage function being tested in https://github.com/containers/podman-desktop-extension-bootc/blob/main/packages/backend/src/build-disk-image.spec.ts

PR #208, opened 2024-03-20
chore(ci): add a workflow that executes e2e tests

odockal:
### What does this PR do?
Introduces E2E test workflow to be run on main branch merge and on demand.
Update: It can also run different fork/repoName/branch using GHA workflow inputs.
#210

### DIFF
--- a/.github/workflows/e2e-main.yaml
+++ b/.github/workflows/e2e-main.yaml
@@ -1,0 +1,119 @@
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: e2e-tests-main
+
+on:
+  push:
+    branches: [main]
+
+  workflow_dispatch:
+    inputs:
+      organization:
+        default: 'containers'
+        description: 'Organization of the Podman Desktop repository'
+        type: string
+        required: true
+      repositoryName:
+        default: 'podman-desktop-extension-bootc'
+        description: 'Podman Desktop Extension BootC repository name'
+        type: string
+        required: true
+      branch:
+        default: 'main'
+        description: 'Podman Desktop Extension BootC repo branch'
+        type: string
+        required: true
+
+jobs:
+  e2e-tests:
+    name: Run E2E tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


odockal:
tested on my fork: https://github.com/odockal/podman-desktop-extension-bootc/actions/runs/8363989131/job/22898241179

benoitf (commented):
hi, why is it not on PR check ?


It looks like it's too late after merge

odockal:
@benoitf If it works, I will add as PR check as well.

cbr7 (approved):

odockal:
> hi, why is it not on PR check ?
> 
> It looks like it's too late after merge

@benoitf We will have more. This is not excluding any other automation beyond merge event. 

I would like to say that on PR check the workflow will behave differently. We should build the extension locally on the workflow, and then use this local image to be installed by the tests. For me this is different effort. Issue here: https://github.com/containers/podman-desktop-extension-bootc/issues/213.

.github/workflows/e2e-main.yaml
```diff
@@ -0,0 +1,119 @@
+#
+# Copyright (C) 2023 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
# Copyright (C) 2024 Red Hat, Inc.
```

.github/workflows/e2e-main.yaml
```diff
@@ -0,0 +1,119 @@
...
+  e2e-tests:
+    name: Run E2E tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.inputs.organization }}/${{ github.event.inputs.repositoryName }}
+          ref: ${{ github.event.inputs.branch }}
+          path: ${{ github.event.inputs.repositoryName }}
+        if: github.event_name == 'workflow_dispatch'
```
benoitf, line 52: for readability, I would move the if as first line

.github/workflows/e2e-main.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+      - uses: actions/checkout@v4
+        with: 
+          path: podman-desktop-extension-bootc
+        if: github.event_name == 'push'
```
benoitf, line 57: move it as first line

.github/workflows/e2e-main.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+      # Checkout podman desktop
+      - uses: actions/checkout@v4
+        with:
+          repository: containers/podman-desktop
+          ref: main
+          path: podman-desktop
```
benoitf, line 64: why do we need to checkout podman desktop ?

odockal: Yes. We need to build repo locally using `yarn test:e2e:build` which build the app in development mode, and then we are spinning up the podman desktop app from there. 
I have not found a way to simply point tests to a podman desktop electron app (binary) and make it run with playwright.

.github/workflows/e2e-main.yaml
odockal: that is right. thx.

.github/workflows/e2e-main.yaml
benoitf: ok b/c yes I think it's not so nice

benoitf (approved):

cdrage (approved):

Issue #149, opened 2024-02-16
Can't build ISO with target arch... even the current one


deboer-tim:
### Bug description

If you try to build an iso with the target arch option, it fails:
```
Generating manifest-iso.json ... WARNING: target-arch is experimental and needs an installed 'qemu-user' package
panic: cannot build iso for different target arches yet

goroutine 1 [running]:
main.cmdBuild(0x400013d700?, {0x40006b6770, 0x1, 0x7})
	/build/bib/cmd/bootc-image-builder/main.go:274 +0x864
github.com/spf13/cobra.(*Command).execute(0x40006bc000, {0x40006b6700, 0x7, 0x7})
	/root/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:983 +0x840
github.com/spf13/cobra.(*Command).ExecuteC(0x4000145b00)
	/root/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
	/root/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039
main.run()
	/build/bib/cmd/bootc-image-builder/main.go:373 +0x530
main.main()
	/build/bib/cmd/bootc-image-builder/main.go:377 +0x1c
```

Unfortunately, it does this for either arch 🤦🏼‍♂️ so you can't build an ISO on 0.3.0.

### Operating system

macOS & windows

### Version

next (development version)

mrguitar:
I'm getting this same error using PD 1.8 and podman machine 5 rc7. I got the same error w/ 4.9. I'm sure I'm doing something wrong, but I don't see a "qemu-user" package in brew. This system does have qemu 8.2.1 installed. I am using the vfkit as the hypervisor. Could that be an issue?

deboer-tim:
Hi @mrguitar. I'm going to work on pulling a new version of bootc image builder where this is fixed... but unfortunately the resulting image still won't boot: https://github.com/osbuild/bootc-image-builder/issues/233. Work on it is active, hopefully we can get this resolved soon.

cdrage:
We could either close this or open a new issue.. right now building .iso works for the current arch (fixed upstream, and our fix here too), but not cross-arch (yet).

deboer-tim:
I opened this really for the 'you can't build with the current arch as target' issue, so I'm going to close.

[...]

Issue #201, opened 2024-03-19
Store build "name" in history json rather than using image / tag / etc.


cdrage:
### Is your enhancement related to a problem? Please describe

Brought up in a meeting,

We should be storing the build name (example, <IMAGE_NAME>-<TAG>-build-1) and anything consecutive afterwards.

Or even have it in the column with "build 1" or something underneath.

### Describe the solution you'd like

Start storing a generated build name for each build

cdrage:
Will require refactor of `history.ts` as we search via tag, arch, etc. selected, instead we should be searching via just the build name.

cdrage:
Assigning this to myself & this sprint as this is a fix for getting multiple parts of the UI working.

We are picking / editing history based upon arch, type, etc... when it should be by name.

PR #223, opened 2024-03-21
chore: build multiple image types (and no target arch api)

### LINKED ISSUE #216: Ability to select multiple image types in one pass

bootc makes it easy to target multiple platforms with a single container image. Yet the plugin makes me create disk images for a single platform at a time which is time consuming for the user.

Maybe changing the radial button to a check box and being able to select multiple image types and have the extension handle queuing them up to BIB on the user's behalf. 

_No response_

### Additional context

"I need a Qcow2, ami, and vmdk. ......tada!!!"  :)

deboer-tim:
### What does this PR do?

Provide the ability to specify multiple image types at a time. Changed the build type from a string to a typed enum array and propagated the change everywhere.

Changing to an enum allowed me to remove the 'Invalid image format' error message, and for now I changed it to loop over all types to check for any existing files for the warning. Changed the radio buttons in the UI to checkboxes to allow multiple selection.

I removed the 'bootc.build.image.location' and 'bootc.build.type' labels from the container (and imagePath parameter to createBuilderImageOptions) since we don't use them, and they don't make sense/would have to be different when you're building multiple image types.

Since it was very closely aligned and would create overlapping PRs, I included the change I started coding this for: changing the API to allow not specifying a target architecture, which causes the --target-arch parameter to not be passed to the cli. No support exists in the frontend, this might be provided later as a fix or temporary workaround for #149.

### Screenshot / video of UI

Screenshot 2024-03-21 at 5 44 54 PM

Fixes #216.

### How to test this PR?

Tests updated. Select multiple image types. Build!


vrothberg (approved):
LGTM, works like a charm, thank you!

cdrage (changes requested):
Screenshot 2024-03-22 at 9 06 40 AM

This is unfortunately causing dashboard issues due to how I added the "addOrUpdate" functionality where it would pick based on type, arch, image, etc:

https://github.com/containers/podman-desktop-extension-bootc/blob/main/packages/backend/src/history.ts#L55

We previously discussed on adding a "build name" for each build.

Let's keep this PR open until we have https://github.com/containers/podman-desktop-extension-bootc/issues/201 fixed

deboer-tim:
Rebased on main. Sorry for using build to do linting :), but there were several one-line changes required after rebasing. @cdrage ready for another review.

cdrage (changes requested):
So only one fix and it's actually unrelated to this PR as it was introduced by the 'id' being changed / added.

When you go to build, it will "timeout" and error out, even though the container was successfully created.

This is because https://github.com/deboer-tim/podman-desktop-extension-bootc/blob/36311c0facc83e57bef0e0558f48034a24968653/packages/frontend/src/Build.svelte#L107 

I had forgotten to also make the change to Build.svelte for searching for history based upon ID rather than arch, type, etc.

Can you change that code to:

```ts
      const found = historyInfo.find(
        info =>
          info.id === buildID,
      );
 ```
 
 Once done, LGTM!

deboer-tim:
Rebased on main due to conflict, made the Build.svelte change and one more in history.ts that I noticed while testing.

cdrage (approved):
Tested, works, LGTM code-wise as well.

Issue #213, opened 2024-03-21
Add E2E tests run as part of PR check


odockal:
### Is your enhancement related to a problem? Please describe

We are proposing new CI workflows in the repo, there is a CI to be run on after merge (https://github.com/containers/podman-desktop-extension-bootc/pull/208). 
To get better and faster results of the changes done in code and while being suggested (new PR), we should have e2e tests run on PR check.

### Describe the solution you'd like

Idea would be to include new steps in PR check workflow (https://github.com/containers/podman-desktop-extension-bootc/blob/main/.github/workflows/pr-check.yaml), similar to what e2e-test-main (once merged) workflow offers, except a couple of adjustments.

It would be ideal to have extension image to be built locally (I guess we cannot really push into registry since we are on PR check and cannot use secrets), then the local image would be used as the one to be installed as an extension in the tests, we can probably use env.vars.  to get image uri.

cbr7:
@deboer-tim @cdrage Hey guys in order to get the e2e tests running on pr check we'd need to be able to build the extension locally on the GHA runner, is the process to how to do that available anywhere?

cdrage:
> @deboer-tim @cdrage Hey guys in order to get the e2e tests running on pr check we'd need to be able to build the extension locally on the GHA runner, is the process to how to do that available anywhere?

Hi! It's on the README:

https://github.com/containers/podman-desktop-extension-bootc?tab=readme-ov-file#contributing

cbr7:
@cdrage great, thank you

PR #224, opened 2024-03-22
chore(tests): initial draft for pr-check e2e tests

cbr7:
### What does this PR do?

Adds e2e test tun to pr-check GHA

https://github.com/containers/podman-desktop-extension-bootc/issues/213


tests/playwright/resources/create-extension.sh
```diff
@@ -0,0 +1,19 @@
...
+cp -r packages/backend/media/ $extensionFolder/media
+cp LICENSE $extensionFolder/
+cp packages/backend/icon-dark.png $extensionFolder/
+cp packages/backend/icon-light.png $extensionFolder/
+cp packages/backend/bootable.woff2 $extensionFolder/
+cp README.md $extensionFolder/
+
+# we need to move the folder with extension content into right podman desktop home folder
+mkdir -p tests/playwright/tests/output/bootc-tests-pd/plugins
+cp -r $extensionFolder tests/playwright/tests/output/bootc-tests-pd/plugins
```
odockal, line 19: it can really be rewritten to a simpler two lines and placed directly into workflow.
```
mkdir -p tests/playwright/tests/output/bootc-tests-pd/plugins/bootcextensionlocal
cd packages/backend && cp -r dist/ package.json media/ LICENSE icon*.png bootable.woff2 README.md ../../tests/playwright/tests/output/bootc-tests-pd/plugins/bootcextensionlocal
```

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,87 @@ jobs:
 
       - name: Run build
         run: yarn build
+  
+  e2e-tests:
+    name: bootc e2e tests
+    runs-on: ubuntu-22.04
+    env:
+      TESTING_ON_PRCHECK: true
```
odockal, line 92: Suggestion: you could use GITHUB_ACTIONS env. var, and/or GITHUB_WORKFLOW which has the name of the workflow to get information from where the test is being run.

odockal (changes requested):

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,84 @@ jobs:
 
       - name: Run build
         run: yarn build
+  
+  e2e-tests:
+    name: bootc e2e tests
```
benoitf, line 89:
```suggestion
    name: e2e tests
```
we're inside the bootc repository anyway

.github/workflows/pr-check.yaml
benoitf: my suggestion here would be to avoid to say how it's run but what action is skipped or done

like SKIP_UNINSTALL, SKIP_SETUP or SKIP_LIFECYCLE so you're not linking the github action to the test

because I could use the same setup by trying to run tests in development mode (and I'm not at this point with a PR check)

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,84 @@ jobs:
...
+            ${{ runner.os }}-yarn-
+
+      - name: Execute yarn in Bootc Extension
+        working-directory: ./podman-desktop-extension-bootc
+        run: yarn --frozen-lockfile
+      
+      - name: Execute the script to build the extension
+        working-directory: ./podman-desktop-extension-bootc
+        run: |
+          yarn
```
benoitf, line 150: why do we need to run yarn again, I don't see a difference with line 145 (same directory)

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,84 @@ jobs:
...
+          # downgrade conmon package version to workaround issue with starting containers, see  https://github.com/containers/conmon/issues/475
+          # remove once the repository contains conmon 2.1.10
+          wget https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_22.04/amd64/conmon_2.1.2~0_amd64.deb -O /tmp/conmon_2.1.2.deb
+          sudo apt install /tmp/conmon_2.1.2.deb
+
+      - name: Build Podman Desktop for E2E tests
+        working-directory: ./podman-desktop
+        run: |
+          yarn --frozen-lockfile
+          yarn test:e2e:build
```
benoitf, line 128: why do we need to build e2e there ?
AFAIK the extension is fetching tests from the 3rd party npmjs repository

cbr7: This builds the podman-desktop in order to be able to be ran and execute the tests from what I understand, am I wrong @odockal ?

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,87 @@ jobs:
...
+          node-version: 20
+      
+      # Checkout podman desktop
+      - uses: actions/checkout@v4
+        with:
+          repository: containers/podman-desktop
+          ref: main
+          path: podman-desktop
+
+      - uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
cbr7: removed

.github/workflows/pr-check.yaml
cbr7: removed.

tests/playwright/resources/create-extension.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+extensionFolder=bootcextensionlocal
+
+yarn
```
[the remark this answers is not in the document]
cbr7: removed it from the script

.github/workflows/pr-check.yaml
cbr7: nice idea, I've switched the environment variable to be named SKIP_INSTALLATION

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,87 @@ jobs:
...
+            ${{ runner.os }}-yarn-
+
+      - name: Execute yarn in Bootc Extension
+        working-directory: ./podman-desktop-extension-bootc
+        run: yarn --frozen-lockfile
+      
+      - name: Execute the script to build the extension
+        working-directory: ./podman-desktop-extension-bootc
+        run: |
+          sudo cp tests/playwright/resources/create-extension.sh ./
```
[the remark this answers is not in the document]
cbr7: I've decided that it's simpler with the script, writing the lines manually doesn't save much since different folders have to be copied in different subfolders so it would end up with a block of command lines.

tests/playwright/package.json
```diff
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@playwright/test": "^1.42.1",
-    "@podman-desktop/tests-playwright": "^1.9.1",
+    "@podman-desktop/tests-playwright": "^0.0.202404081045-e2142a7",
```
benoitf, line 13: why do we need to use a non-tagged version there ?

cbr7: I'm not positive about 1.9.1 specifically, however we make changes to the test framework all the time so tagged version very quickly become out of data from that perspective.

.github/workflows/pr-check.yaml
odockal: @benoitf @cbr7 that seems correct. We need to build podman desktop in a development mode locally so we can launch it from e2e tests using `electron.launch(args:['path/to/podman-desktop'])`.

tests/playwright/package.json
benoitf: but then if you always want to grab the current "snapshot" of Podman Desktop, probably you should update the version during the e2e test else for example in a couple of days, it'll the same issue

Either we stick to the version being supported or if you need always the latest 'bits' then the test should take care of switching the dependency to the current published version of the test framework

or maybe two run are interesting.
Test against the latest stable version of Podman Desktop and test against the current development version of Podman Desktop

cbr7: After some advice from @odockal I've changed the npm package version for the test framework to have value `next`, this will ensure that the latest version of the test framework is used when running the tests and removes the need to manually update the version in the package.json file.

benoitf: so it won't do anything to have next

it's b/c we're pinning down versions into a yarn.lock file so versions are resolved only once (when we submit the yarn.lock file)

cbr7: ah yes, fair point, in that case it doesn't do anything.

odockal: Unless we would remove the `@podman-desktop/tests-playwright` from the lock file, or we remove the file on PR check to make sure we always grab latest version.

benoitf: in PR check yaml, do `yarn add -D @podman-desktop/tests-playwright@next` in tests/playwright folder as first step

cbr7: done, added it before the lockfile is ran.

cbr7:
@deboer-tim @cdrage @odockal I think this PR is finally ready, if you guys can take a look and approve we can merge it

tests/playwright/resources/create-extension.sh
```diff
@@ -0,0 +1,36 @@
...
+yarn build
+mkdir -p $extensionFolder
+cp -r packages/backend/dist/ $extensionFolder/dist
+cp packages/backend/package.json $extensionFolder/
+cp -r packages/backend/media/ $extensionFolder/media
+cp LICENSE $extensionFolder/
+cp packages/backend/icon-dark.png $extensionFolder/
+cp packages/backend/icon-light.png $extensionFolder/
+cp packages/backend/bootable.woff2 $extensionFolder/
+cp README.md $extensionFolder/
```
benoitf, line 32: I don't really know what it does but we should not do this because it'll lead to errors, unmaintained stuff, etc.

The extension has a Containerfile that builds the extension.

if we need to have a 'build-ready' folder of the extension, it should reuse that Containerfile has it's the own that will package the extension.

Else for example you may have omitted here some files that are included correctly in the release or you may have files that are there but not in the Containerfile so PR will pass but end users will not have the extension working

to me e2e tests are to really tests what is being 'produced' so it should reuse the build part of the extension

tests/playwright/src/bootc-extension.spec.ts
```diff
@@ -41,13 +41,14 @@ const imageName = 'quay.io/centos-bootc/fedora-bootc';
 const containerFilePath = path.resolve(__dirname, '..', 'resources', 'bootable-containerfile');
 const contextDirectory = path.resolve(__dirname, '..', 'resources');
 const isLinux = os.platform() === 'linux';
+const cicd = process.env.SKIP_INSTALLATION;
```
benoitf, line 44: cicd is not really helpful based on the env var

please rename it to `skipInstallation`

benoitf (changes requested):

tests/playwright/src/bootc-extension.spec.ts
cbr7: done

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,87 @@ jobs:
...
+
+      - uses: actions/cache@v4
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Ensure latest version of test  framework
```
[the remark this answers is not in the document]
cbr7: done

tests/playwright/resources/create-extension.sh
cbr7: @benoitf The script represents how to build the extension locally, given that it's supposed to run on each individual branch, if you have another solution of how that can be achieved that would be great.

odockal: I agree. 
Although, having extension built locally (and tagged) using containerfile will not help us either, since then I have problem with authentication of the local repository when using locally built image as an input for install extension from OCI image. So we do not have a valid image from upstream on pr check (we might consider adding that, but then we are facing problem with secrets... )

Do you suggest something like this then https://github.com/odockal/podman-desktop-qe/blob/main/scripts/extension-installation-local.sh#L37-L53 ?

Otherwise I do not know how to do it.

benoitf: so I still believe we should try to install the image locally built to really test what we're building and how users will handle the image
but we can 'postpone' the issue you seem to have with authentication

as intermediate step, yes let's copy the content of the image

1. build the image
2. create the container like `CONTAINER_ID=$(podman create my-local-bootc-image)`
3. dump the content `podman export $CONTAINER_ID > /tmp/content.tar`
4. extract/save the content to `~/.local/share/containers/podman-desktop/plugins` folder (as export will have an extension folder, we will have ``~/.local/share/containers/podman-desktop/plugins/extension` once extracted

and let's replace soon that by installing the extension from Podman Desktop itself

benoitf: and no need to use `--extension-folder` arg when launching Podman Desktop

benoitf: also we should launch Podman Desktop binary (the one of production mode) not in development mode but I think here it's a 'known issue' about being able to run electron + playwright without dev mode

but at least we should be testing the 'production bits' of the extension, not the development mode

odockal: @benoitf @cbr7 That sounds doable then! Thanks!

benoitf: also could you say what is the issue if you publish the image to a local registry 

AFAIK this one is already fixed: https://github.com/containers/podman-desktop/issues/6344

benoitf: as I guess it would be better if we could publish to localhost:500 registry and install the extension from that registry as part of the workflow

odockal: @benoitf the error says:
```
Error while analyzing image: Error: Unable to find auth info for https://localhost/v2/. Error: RequestError
```

which might imply that the image itself is not part of any registry? 
I am simply building/tagging the image locally. no push into local registry. That might be a requirement?

benoitf: The extension image is pulled by Podman Desktop itself (not by podman). localhost for Podman Desktop means your laptop, not the vm where podman machine might run. And it's discussing through registry protocol, so it'll call `/v2` path on the path to interact with.

so yes, if you want to access something being built by Podman, image needs to be exposed to a registry first.

odockal: @benoitf oh, alright, thanks for explanation. I am trying to set it up locally, so we could automate in the workflow, but I am not really successful atm (I tried what you are proposing in the #6344, though I got `panic: unable to configure authorization (htpasswd): open /auth/htpasswd: permission denied`). I would propose that we do now `build+export+extract` path how to install the extension, and then create a follow up issue for `building + pushing + installing` the extension via oci image from local registry, which gonna require some more testing. wdyt?

benoitf: sure one problem after another one

most important one is to use the artifact being produced ( consuming it from the OCI registry is another step)

odockal: Perfect!

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,91 @@ jobs:
...
+      - name: Build bootc extension from container file
+        working-directory: ./podman-desktop-extension-bootc
+        run: |
+          yarn build
+          podman build -t local_bootc_image ./
+          CONTAINER_ID=$(podman create localhost/local_bootc_image)
+          podman export $CONTAINER_ID > /tmp/local_bootc_extension.tar
+          sudo cp /tmp/local_bootc_extension.tar ~/.local/share/containers/podman-desktop/plugins
+          cd ~/.local/share/containers/podman-desktop/plugins/
+          tar -xvf local_bootc_extension.tar
```
benoitf, line 160: use `-C ~/.local/share/containers/podman-desktop/plugins/` to tar command ?

tests/playwright/resources/create-extension.sh
```diff
@@ -0,0 +1,36 @@
+: '/**********************************************************************
```
benoitf, line 1: remove the file now ?

.github/workflows/pr-check.yaml
```diff
@@ -84,3 +84,92 @@ jobs:
...
+        working-directory: ./podman-desktop-extension-bootc
+        run: |
+          yarn build
+          podman build -t local_bootc_image ./
+          CONTAINER_ID=$(podman create localhost/local_bootc_image --entrypoint "")
+          podman export $CONTAINER_ID > /tmp/local_bootc_extension.tar
+          mkdir -p ~/.local/share/containers/podman-desktop/plugins/
+          sudo cp /tmp/local_bootc_extension.tar ~/.local/share/containers/podman-desktop/plugins/
+          cd ~/.local/share/containers/podman-desktop/plugins/
+          tar -xvf local_bootc_extension.tar
```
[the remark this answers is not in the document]
cbr7: done

benoitf (approved):

tests/playwright/resources/create-extension.sh
cbr7: Should be good now, the extension is built from the container file and extracted into the plugin folder to be picked up when PD starts up.

.github/workflows/pr-check.yaml
odockal: Just the location needs to be one used by e2e tests. This is for podman installed locally. E2E tests use different location. The `-C` flag is correct though!

tests/playwright/resources/create-extension.sh
odockal: Follow up issue: https://github.com/containers/podman-desktop-extension-bootc/issues/330

odockal:
LGTM! Good work @cbr7 and great review @benoitf

PR #229, opened 2024-03-22
nit: Homepage not Dashboard

cdrage:
nit: Homepage not Dashboard

### What does this PR do?

Should be Homepage not Dashboard for the breadcrumb / comments

### Screenshot / video of UI

Screenshot 2024-03-22 at 3 13 45 PM

N/A

### How to test this PR?

You'll see it say Homepage not Dashboard when clicking on Build.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -92,7 +92,7 @@ async function buildBootcImage() {
     bootcClient.buildImage(buildOptions);
 
     // Continue doing listHistoryInfo until the build container name, tag, type and arch show up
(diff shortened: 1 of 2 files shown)


benoitf (approved):

deboer-tim (changes requested):
We call it the 'homepage' in the issue/PR/code, but when users opens it the title is 'Bootable Containers'. In PD the breadcrumb always lists the previous page's title, so I actually think breadcrumb.ts:23 should say 'Bootable Containers'.

While we're at it, breadcrumb.ts:23 is where it's currently being picked up, but the reference in App.svelte:35 is where breadcrumbs are normally picked up and what goes to telemetry, so can you update it there too?

PR #230, opened 2024-03-22
nit: fix naming to match extension

cdrage:
nit: fix naming to match extension

### What does this PR do?

Fixes the naming BootC to Bootable Container like the rest of the
extension.

This includes any breadcrumbs as well.

N/A

### How to test this PR?

Build / view the pages.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -157,7 +157,7 @@ $: {
 }
 </script>
 
-<FormPage title="BootC Image Builder" inProgress="{buildInProgress}" showBreadcrumb="{true}">
+<FormPage title="Build Disk Image" inProgress="{buildInProgress}" showBreadcrumb="{true}">
   <svelte:fragment slot="icon">
     <i class="fas fa-rocket fa-2x" aria-hidden="true"></i>
   </svelte:fragment>
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -20,10 +20,10 @@ onMount(() => {
 });
 </script>
 
-<Route path="/*" breadcrumb="Home" isAppMounted="{isMounted}" let:meta>
+<Route path="/*" breadcrumb="Bootable Containers" isAppMounted="{isMounted}" let:meta>
   <main class="flex flex-col w-screen h-screen overflow-hidden bg-charcoal-700">
     <div class="flex flex-row w-full h-full overflow-hidden">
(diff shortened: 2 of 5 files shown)


packages/frontend/src/Build.svelte
```diff
@@ -157,7 +157,7 @@ $: {
 }
 </script>
 
-<FormPage title="BootC Image Builder" inProgress="{buildInProgress}" showBreadcrumb="{true}">
```
deboer-tim, line 160: In the rest of PD we (mostly) have a pattern where the breadcrumb is the same as the action that got you there (or slightly longer), and the title is just a more verbose or specific version of that, e.g.:
```
action=Deploy to Kubernetes
breadcrumb=Deploy to Kubernetes
title=Deploy generated pod to Kubernetes
```
or
```
action=Pull
breadcrumb=Pull an Image
title=Pull image from a registry
```

So today we have:
```
action=Build Disk Image (or Build from the homepage)
breadcrumb=BootC Image Builder
title=BootC Image Builder
```

What do you think about keeping this pattern? To me it helps to be consistent, and keeps it to 'what the user is trying to do' and not 'what underlying tool are we using':
```
action=Build Disk Image (or Build from the homepage)
breadcrumb=Build Disk Image
title=Build a disk image
or title=Build a disk image from container image
```

cdrage:
This PR replaces #229 

Also updates it so it looks like this when going to build:

Screenshot 2024-03-22 at 3 34 23 PM

packages/frontend/src/lib/BootcEmptyScreen.svelte
```diff
@@ -5,5 +5,5 @@ import BootcIcon from './BootcIcon.svelte';
 
 <EmptyScreen
   icon="{BootcIcon}"
-  title="No bootable container builds found"
```
deboer-tim, line 8: I think the lower case was actually correct here. Bootable container is not a trademark or title for something, it's just a name like 'pod',  'image' or 'container' where we don't capitalize.

Unfortunate that we need to show builds here, otherwise we could call these disk images. Or maybe we should shift and start making this page more like the others and just show disk images??

deboer-tim (changes requested):

cdrage:
Updated with the changes, I agree with keeping the same pattern and I followed your comment, but may have gotten some parts wrong, let me know if that's good!

deboer-tim (approved):

PR #231, opened 2024-03-22
chore: create unique entries in history when building

cdrage:
chore: create unique entries in history when building

### What does this PR do?

* Instead of overriding the previous history, create a unique entry
  using name
* Adds 'name' to the interface

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/dc2ba8f1-a04b-4219-8edc-0777b45eabf1

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/201

### How to test this PR?

1. Delete `rm
   ~/.local/share/containers/podman-desktop/extensions-storage/redhat.bootc/history.json`
2. Create multiple builds, they should appear now as multiple builds.

### DIFF
--- a/packages/backend/src/build-disk-image.ts
+++ b/packages/backend/src/build-disk-image.ts
@@ -34,7 +34,7 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
   let errorMessage: string;
 
   const requiredFields = [
-    { field: 'name', message: 'Bootc image name is required.' },
+    { field: 'id', message: 'Bootc image id is required.' },
     { field: 'tag', message: 'Bootc image tag is required.' },
     { field: 'type', message: 'Bootc image type is required.' },
     { field: 'engineId', message: 'Bootc image engineId is required.' },
@@ -100,12 +100,13 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
   // "Returning" withProgress allows PD to handle the task in the background with building.
   return extensionApi.window
     .withProgress(
-      { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + build.name },
+      { location: extensionApi.ProgressLocation.TASK_WIDGET, title: `Building disk image ${build.image}` },
       async progress => {
-        const buildContainerName = build.name.split('/').pop() + bootcImageBuilderContainerName;
+        const buildContainerName = build.image.split('/').pop() + bootcImageBuilderContainerName;
         let successful: boolean = false;
         let logData: string = 'Build Image Log --------\n';
-        logData += 'Image:  ' + build.name + '\n';
+        logData += 'ID:     ' + build.id + '\n';
+        logData += 'Image:  ' + build.image + '\n';
         logData += 'Type:   ' + build.type + '\n';
         logData += 'Folder: ' + build.folder + '\n';
         logData += '----------\n';
@@ -125,7 +126,7 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
         const containerName = await getUnusedName(buildContainerName);
         const buildImageContainer = createBuilderImageOptions(
           containerName,
-          `${build.name}:${build.tag}`,
+          `${build.image}:${build.tag}`,
           build.type,
           build.arch,
           build.folder,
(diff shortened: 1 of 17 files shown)


cdrage:
Note this is a **BREAKING** change.

In the next release we should mention that you must delete: `~/.local/share/containers/podman-desktop/extensions-storage/redhat.bootc/history.json`

This is because we are now adding 'name' to the history.json

benoitf:
Can we detect the old file and migrate it? So no breaking change

benoitf:
or just delete it if it has invalid / old format

deboer-tim:
> or just delete it if it has invalid / old format

Yes, that's what I was thinking as well. Limited users so far, and the only purpose it had was pre-filling things like folder names. Not a huge loss if we just start fresh.

cdrage:
Ready for review 🎉 

I have added a function as well that will check to see the old history.json format if there is any missing 'name'.

Instead of deleting the old file, I instead just had it add the 'name' entry to it since it was easy enough to do. Keeping compatibility / history for old versions of this extension.

deboer-tim (changes requested):
Had a side chat with Charlie. The generated names help us refer to a unique disk image generation, but they don't really help the user (no more meaningful that the other columns) and weren't specified by the user.

I think we should just treat the names like ids (or even rename to id) and remove the Name column so that they aren't exposed. It solves the underlying problem and keeps the table more concise.

cdrage:
> Had a side chat with Charlie. The generated names help us refer to a unique disk image generation, but they don't really help the user (no more meaningful that the other columns) and weren't specified by the user.
> 
> I think we should just treat the names like ids (or even rename to id) and remove the Name column so that they aren't exposed. It solves the underlying problem and keeps the table more concise.

Updated to now use `id` for the "unique name" and removed the column from the view. This means we can easily update + build multiple images without an issue.

packages/backend/src/build-disk-image.ts
```diff
@@ -100,12 +100,13 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
   // "Returning" withProgress allows PD to handle the task in the background with building.
   return extensionApi.window
     .withProgress(
-      { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + build.name },
+      { location: extensionApi.ProgressLocation.TASK_WIDGET, title: 'Building disk image ' + build.id },
```
deboer-tim, line 103: I think we should keep this one to build.image (or build.image+":"+build.tag). If I launch it several days in a row I'd expect the tasks to still say it's building "image-name", not "image-name 5".

packages/backend/src/build-disk-image.ts
```diff
@@ -100,12 +100,13 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
...
       async progress => {
-        const buildContainerName = build.name.split('/').pop() + bootcImageBuilderContainerName;
+        const buildContainerName = build.id + bootcImageBuilderContainerName;
```
deboer-tim, line 105: Ok to merge, but ditto here or maybe we find a non-conflicting name later. If there's no current container with a conflict I'd expect the name to be "image-name".

deboer-tim (approved):
Thanks for the changes. One minor thing I'd like you to fix, but otherwise ready to merge.

cdrage:
Fixed! Thanks for the feedback.

Confirmed as well with building that this works / changes are now present in json / etc.
Screenshot 2024-03-26 at 5 34 02 PM


```
cat /Users/cdrage/.local/share/containers/podman-desktop/extensions-storage/redhat.bootc/history.json
[
  {
    "id": "test-2",
    "image": "quay.io/podai/test",
    "tag": "latest",
    "engineId": "podman.Podman Machine",
    "folder": "/Users/cdrage/bootc123",
    "type": "raw",
    "arch": "amd64",
    "status": "running",
    "timestamp": "2024-03-26T21:33:53.287Z",
    "buildContainerId": "b578d2625e41b8f2319b9ada5e9f14bfd67d6f189b48ae7513ef72c32ee5421a"
  },
  {
    "id": "test",
    "image": "quay.io/podai/test",
    "tag": "latest",
    "engineId": "podman.Podman Machine",
    "folder": "/Users/cdrage/bootc",
    "type": "raw",
    "arch": "amd64",
    "status": "running",
    "timestamp": "2024-03-26T21:33:47.998Z",
    "buildContainerId": "e818b90ca90b80b0f3a04c5c94fcf1f6eec8bb16199291be1070b3fefbda679b"
  }
]%
```

PR #233, opened 2024-03-25
chore: fix end status

### LINKED ISSUE #232: Successful message should not appear until after build is complete

The dialog to say that the build task is successful appears before the task is actually finished.

The job should be complete before/as the dialog appears.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Fixing two minor issues related to the end of the build task:

- The bug in PD that required us to report an increment of -1 to trigger task completion appears to have been fixed a while ago. Better to report 100%.

- Currently, the dialog to tell you that the job is successful blocks the task from finishing until you say OK. It should be async.

### Screenshot / video of UI

N/A

Fixes #232.


cdrage:
Hard to make it from the formatting, but was able to see where it was changed and it LGTM. Tested and was able to still build / increment / dialog working.

PR #235, opened 2024-03-25
feat: open webview from image action

deboer-tim:
### What does this PR do?

Removes the quick pick command, and instead triggers opening the webview, switching to
the build page, and preselecting the image.

### Screenshot / video of UI

N/A

### How to test this PR?

Tests still to-do.


slemeur:
Is this needed?

deboer-tim:
This is removing the quick pick and making all the paths go to the webview - isn't that the intention?

packages/frontend/src/App.svelte
```diff
@@ -26,6 +33,9 @@ onMount(() => {
       <Route path="/" breadcrumb="Bootable Containers">
         <Homepage />
       </Route>
+      <Route path="/build2/:name/:tag" breadcrumb="Build" let:meta>
```
cdrage, line 36: What we could do instead is modify it to `<Route path="/build/:name/:tag" breadcrumb="Build" let:meta>` / have the same route.

And then within Build.svelte if name / tag is blank, do the normal current functionality.

deboer-tim:
Added tests, and rebased on main. Ready for full review.

packages/backend/src/extension.ts
```diff
@@ -106,6 +81,40 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
+      await openBuildPage(panel, image);
+    }),
+  );
+}
+
+export async function openBuildPage(
+  panel: extensionApi.WebviewPanel,
+  image: { name: string; tag: string },
+): Promise<void> {
+  console.log('Opening webview for ' + image.name);
```
cdrage, line 96: Remove log?

packages/backend/src/extension.ts
```diff
@@ -106,6 +81,40 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
+  if (!bootcWebView) {
+    console.error('Could not find bootc webview');
+    return;
+  }
+
+  await extensionApi.navigation.navigateToWebview(bootcWebView.id);
+
+  // if we trigger immediately, the webview hasn't loaded yet and can't redirect
+  // if we trigger too slow, there's a visible flash as the homepage appears first
+  await new Promise(r => setTimeout(r, 100));
```
cdrage, line 111: Unfortunate that we're unable to figure out if the extension has loaded yet before switching to the web view. 

Has there been a solution for this within ai studio or podman desktop / they've encountered the same thing when redirecting after loading the web view?

@feloy @axel7083 @lstocchi

packages/backend/src/extension.ts
```diff
@@ -106,6 +81,40 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
+
+  await panel.webview.postMessage({
+    id: Messages.MSG_NAVIGATE_BUILD,
+    // Must pass in an empty body to satisfy the type system. If it is undefined, this fails
```
cdrage, line 115: can remove this comment

packages/frontend/src/App.svelte
```diff
@@ -17,6 +19,11 @@ onMount(() => {
   const state = getRouterState();
   router.goto(state.url);
   isMounted = true;
+
+  return rpcBrowser.subscribe(Messages.MSG_NAVIGATE_BUILD, (x: string) => {
+    console.log(`Navigating to /build/${x}`);
```
cdrage, line 24: remove log?

packages/frontend/src/Build.svelte
```diff
@@ -23,28 +26,46 @@ let buildFolder: string;
...
 
-    // Find the image that matches the latest build's name and tag and set selectedImage to that value
-    selectedImage = bootcAvailableImages.find(
-      image =>
-        image.RepoTags && image.RepoTags.length > 0 && image.RepoTags[0] === `${latestBuild.image}:${latestBuild.tag}`,
-    );
+  // If an image name and tag were passed in, try to use it as the initially selected image
+  let initialImage: ImageInfo | undefined;
+  if (imageName && imageTag) {
+    console.log('Preselecting image: ' + imageName + ' ' + imageTag);
```
cdrage, line 56: remove log?

packages/frontend/src/Build.svelte
```diff
@@ -23,28 +26,46 @@ let buildFolder: string;
...
+    initialImage = findImage(`${imageName}:${imageTag}`);
+  }
+
+  // If not, use the last image from history if it is valid
+  if (!initialImage && historyInfo.length > 0) {
+    const latestBuild = historyInfo[0];
+    // Find the image that matches the latest build's name and tag
+    initialImage = findImage(`${latestBuild.image}:${latestBuild.tag}`);
```
cdrage, line 64: should we check if .image + .tag isn't blank / undefined?

cdrage (changes requested):

packages/backend/src/extension.ts
deboer-tim: Sure, fixed in next commit.

packages/backend/src/extension.ts
deboer-tim: Sure, fixed in next commit.

packages/frontend/src/App.svelte
deboer-tim: Sure, fixed in next commit.

packages/frontend/src/Build.svelte
deboer-tim: Well, we weren't before and it'll just fail silently :) but I check and simplified this in next commit.

packages/frontend/src/Build.svelte
deboer-tim: Sure, fixed in next commit.

cdrage (approved):
Approved, but let's open an issue for refactoring / fixing the bug with `await new Promise(r => setTimeout(r, 100));`

PR #246, opened 2024-03-28
chore: temporarily ignore arch when building iso

deboer-tim:
### What does this PR do?

Image builder has had an issue for a while now where if you specify the target arch when building an iso it will fail, even if you're specifying the current/default arch. We've had enough users hit this (so this extension is the 'blocking' issue) even though the generated images are also broken/unusable.

We've discussed possible UI changes (like having a 'no target arch' option, or allowing no arch to be selected) but we don't really like these and do not want to make UI changes that could be very temporary - so this is the simplest workaround for now that will unblock users.

Once the following bug is resolved we will update image builder, remove this workaround, and if necessary change the UI.

https://github.com/osbuild/bootc-image-builder/issues/316

### Screenshot / video of UI

N/A

Temporary fix for #149.

### How to test this PR?

Build an iso and confirm image builder doesn't fail.

### DIFF
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -40,7 +40,7 @@ beforeEach(() => {
 
 test('check image builder options', async () => {
   const image = 'test-image';
-  const type = 'iso';
+  const type = 'raw';
   const arch = 'amd';
   const name = 'my-image';
   const outputFolder = '/output-folder';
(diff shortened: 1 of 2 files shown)


cdrage (approved):
Tested with `iso` and was able to (finally!) build. LGTM 👍 

Thanks for the test too!


https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/88fd51b9-df91-4a74-99df-4bc802226ee7

PR #259, opened 2024-03-29
chore(README): Root pw clarification, link other docs, drop -dev

cgwalters:
- The root password I think is dangerous *especially* for anything involving uploading an AMI etc.  As the comment says it's only not catastrophic because of `PermitRootLogin prohibit-password` in our default images.  To really make this better we need to streamline the "launch with local virt" flow, which I think is best done with integration with podman-bootc which needs further work.
- Emphasize the difference between disk images and container images across the board more
- Link to new bootc generic docs on things like users/groups
- No need to use the `-dev` images

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,12 @@ Easily go from container to VM / ISO-on-a-USB / RAW image!
 
 ## Technology
 
-The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container OS images. 
+The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container disk images which derive from
+a bootable input *container* image. The "source of truth" in general is the container image,
+and the disk images are derived from it. Once a machine is created from the disk image, it
+can apply transactional updates "in place" from newly pushed container images (without creating a new disk image).
 
-Within [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) the tool uses [bootc](https://containers.github.io/bootc/) as a basis for conversion to achieve the bootable container OS as well as libraries such as [libostree](https://ostreedev.github.io/ostree/) and [ostree-rs-ext](https://github.com/ostreedev/ostree-rs-ext).
+For more, see [bootc](https://containers.github.io/bootc/).
 
 The **ONLY** currently supported base image is [`quay.io/centos-bootc/fedora-bootc`](https://centos.github.io/centos-bootc). More are to be supported in the future!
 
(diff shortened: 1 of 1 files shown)


cdrage (approved):

README.md
```diff
@@ -17,9 +17,12 @@ Easily go from container to VM / ISO-on-a-USB / RAW image!
 
 ## Technology
 
-The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container OS images. 
+The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container disk images which derive from
+a bootable input *container* image.   The "source of truth" in general is the container image,
```
benoitf, line 21:
```suggestion
a bootable input *container* image. The "source of truth" in general is the container image,
```

README.md
```diff
@@ -40,14 +43,22 @@ You are "creating" an OS straight from a Containerfile, isn't that awesome?
 Want a quick straight-to-the-point Hello World Containerfile?
 
 ```Dockerfile
-FROM quay.io/centos-bootc/centos-bootc-dev:stream9
-# Change your root password for a "test login"
+FROM quay.io/centos-bootc/centos-bootc:stream9
+# Change your root password for a "test login" that
+# allows to log in on a virtual/physical console
+# NOTE: While some base images may set `PermitRootLogin prohibit-password`
+# for OpenSSH, not all will.  Those that don't make this pattern
```
benoitf, line 50:
```suggestion
# for OpenSSH, not all will. Those that don't make this pattern
```

benoitf (changes requested):
Could you please add a short title indicating using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) what the PR does rather than 'A lot of changes' that could be like on all PRs

There are multiple non-required spaces in the README

README.md
```diff
@@ -17,9 +17,12 @@ Easily go from container to VM / ISO-on-a-USB / RAW image!
 
 ## Technology
 
-The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container OS images. 
+The **Bootable Container (bootc)** extension utilizes [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) in order to create bootable container disk images which derive from
+a bootable input *container* image.   The "source of truth" in general is the container image,
+and the disk images are derived from it.  Once a machine is created from the disk image, it
```
benoitf, line 22:
```suggestion
and the disk images are derived from it. Once a machine is created from the disk image, it
```

cgwalters:
> Could you please add a short title indicating using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) 

My commit messages are "Linux kernel style" which is definitely a convention, but yes - I should have run "git log" to see.  

> what the PR does rather than 'A lot of changes' that could be like on all PRs

Hard to summarize in a few characters, the commit message has things broken down, but I tried with the next push.

> There are multiple non-required spaces in the README

Heh, I got called out for that before, I'm just old enough that it was actually kind of a convention...but yes, definitely not anymore.  I don't personally notice the spaces either way.

benoitf (approved):
thanks @cgwalters

PR #261, opened 2024-04-02
feat: error if running on older podman version

### LINKED ISSUE #168: Should check if podman 5.0 or not

5.0 is a requirement and we should be checking for that. 

Fail if below 5.0 since we cannot build multi-arch on mac m1 without 5.0

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Provide a function to test if the machine we're running on is Podman v5 or higher, and use it to trigger an error if we try running an image build on an older machine.

### Screenshot / video of UI

N/A

Fixes #168.

### How to test this PR?

Automated tests; try running a build with a <5.0 machine.

### DIFF
--- a/packages/backend/src/build-disk-image.ts
+++ b/packages/backend/src/build-disk-image.ts
@@ -58,6 +58,13 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
     throw new Error('The podman machine is not set as rootful.');
   }
 
+  const isPodmanV5 = await machineUtils.isPodmanV5Machine();
+  if (!isPodmanV5) {
+    const errorMessage = 'Podman v5.0 or higher is required to build disk images.';
+    await extensionApi.window.showErrorMessage(errorMessage);
+    throw new Error('The podman machine version is below v5.0.');
+  }
+
   // Use build.type to check for existing files
   let promptForOverwrite: boolean = false;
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/build-disk-image.ts
```diff
@@ -58,6 +58,13 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
     throw new Error('The podman machine is not set as rootful.');
   }
 
+  const isPodmanV5 = await machineUtils.isPodmanV5Machine();
+  if (!isPodmanV5) {
+    const errorMessage = 'Podman v5.0 or higher is required to build disk images.';
+    await extensionApi.window.showErrorMessage(errorMessage);
+    throw new Error('The podman machine version is below v5.0.');
```
cdrage, line 65:
```suggestion
    throw new Error('Podman Machine is below v5.0');
```

Small syntax suggestion

cdrage (approved):
Tested with 4.9.3 and the error appears (as normal).

LGTM

packages/backend/src/build-disk-image.ts
deboer-tim: Fixed in new commit.

packages/backend/src/machine-utils.spec.ts
```diff
@@ -158,3 +158,59 @@ test('Pass true if Rootful is in HostUser', async () => {
...
+    return true;
+  });
+
+  // Mock the readFile function to return the "fake" file with rootful being true
+  const spyReadFile = vi.spyOn(fs.promises, 'readFile');
+
+  // Mock reading the file to have Rootful as true
+  spyReadFile.mockResolvedValue(JSON.stringify({ Rootful: true }));
+  await expect(machineUtils.isPodmanV5Machine()).resolves.toBe(true);
+});
```
benoitf, line 216: when I see all this mock just to return a dummy value like '4.0.0' or '5.0.0', it makes me feel that mocking would be so simple using class/objects rather than functions like getMachineInfo

Issue #245, opened 2024-03-28
Auto-detect arch of image, do not select different arch


cdrage:
### Is your enhancement related to a problem? Please describe

Why we cannot select multiple arch's / but we can display them:
* listImages API / PD can only list images right now, not manifests (manifests contain multi-arch images).
* if the image is amd64, you can only build amd64... etc. do not select the other architectures.

### Describe the solution you'd like

When selecting an image, detect what architecture it is, and only select the one that matches the architecture.

cdrage:
FYI, let's wait until https://github.com/containers/podman-desktop/issues/6236 is completed, so we may use the new API calls such as inspectManifest, createManifest, listing, etc in order to correctly inspect the image and determine what architecture they are using.

PR #264, opened 2024-04-02
chore: add inspectImage API

cdrage:
chore: add inspectImage API

### What does this PR do?

Adds the inspectImage API for the frontend so that we can inspect the
image for the architecture / OS / more information before building.

Part of https://github.com/containers/podman-desktop-extension-bootc/issues/245

### Screenshot / video of UI

N/A

Part of https://github.com/containers/podman-desktop-extension-bootc/issues/245

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/339

### How to test this PR?

Covered by tests

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -113,6 +113,19 @@ export class BootcApiImpl implements BootcApi {
     return images;
   }
 
+  async inspectImage(image: ImageInfo): Promise<podmanDesktopApi.ImageInspectInfo> {
+    let imageInspect: podmanDesktopApi.ImageInspectInfo;
+    try {
+      imageInspect = await podmanDesktopApi.containerEngine.getImageInspect(image.engineId, image.Id);
+    } catch (err) {
+      throw new Error(`Error inspecting image: ${err}`);
+    }
+    if (imageInspect === undefined) {
+      throw new Error('Unable to retrieve image inspect information');
+    }
+    return imageInspect;
+  }
+
   async listHistoryInfo(): Promise<BootcBuildInfo[]> {
     try {
       // Load the file so it retrieves the latest information.
--- a/packages/backend/src/container-utils.ts
+++ b/packages/backend/src/container-utils.ts
@@ -45,6 +45,17 @@ export async function getContainerEngine(): Promise<extensionApi.ContainerProvid
   return runningPodmanConnections[0].connection;
 }
 
+// Inspect the image / get more information
+export async function inspectImage(engineId: string, image: string): Promise<extensionApi.ImageInspectInfo> {
+  console.log('Inspecting image: ', image);
+  try {
+    return await extensionApi.containerEngine.getImageInspect(engineId, image);
+  } catch (e) {
+    console.error(e);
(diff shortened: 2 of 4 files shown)


packages/backend/src/api-impl.ts
```diff
@@ -103,6 +103,16 @@ export class BootcApiImpl implements BootcApi {
...
   }
 
+  async inspectImage(image: ImageInfo): Promise<podmanDesktopApi.ImageInspectInfo> {
+    try {
+      return await podmanDesktopApi.containerEngine.getImageInspect(image.engineId, image.Id);
+    } catch (err) {
+      await podmanDesktopApi.window.showErrorMessage(`Error inspecting image: ${err}`);
+      console.error('Error inspecting image: ', err);
+    }
+    return {} as podmanDesktopApi.ImageInspectInfo;
```
benoitf, line 113: I'm not sure to understand the design like return await and return {} at the end ?

packages/backend/src/container-utils.ts
```diff
@@ -44,6 +44,17 @@ export async function getContainerEngine(): Promise<extensionApi.ContainerProvid
   return runningPodmanConnections[0].connection;
 }
 
+// Inspect the image / get more information
+export async function inspectImage(engineId: string, image: string): Promise<extensionApi.ImageInspectInfo> {
+  console.log('Inspecting image: ', image);
```
benoitf, line 49: to be removed

packages/backend/src/api-impl.ts
```diff
@@ -103,6 +103,16 @@ export class BootcApiImpl implements BootcApi {
     return images;
   }
 
+  async inspectImage(image: ImageInfo): Promise<podmanDesktopApi.ImageInspectInfo> {
+    try {
+      return await podmanDesktopApi.containerEngine.getImageInspect(image.engineId, image.Id);
+    } catch (err) {
+      await podmanDesktopApi.window.showErrorMessage(`Error inspecting image: ${err}`);
```
deboer-tim, line 111: If this fails we shouldn't pop up error messages from main. The caller (likely frontend in this case) should decide the user behaviour when you can't get the info.

packages/backend/src/container-utils.ts
cdrage: added to match the other parts of container utils

packages/backend/src/api-impl.ts
deboer-tim: Ditto, assuming this is for the error case and should be like a rejected promise returned to the frontend.

cdrage: updated!

deboer-tim (changes requested):

packages/backend/src/api-impl.ts
benoitf: Usually pattern is throw Error or Promise<Object | undefined> not an empty object cast as the expected result

cdrage:
Ready for review 💯 @deboer-tim

deboer-tim (approved):
LGTM

Issue #247, opened 2024-03-28
Github & dev tidy-up


deboer-tim:
### Is your enhancement related to a problem? Please describe

Briefly came up during a chat - there are a few really minor things that we should do to confirm github/codebase is all clean and working.

### Describe the solution you'd like

- Confirm dependabot is working/latest versions of dependencies.
- `yarn test` should work from root folder (just needs a rename from `yarn test:unit`)
- Quick compare with other extensions to see if there are any more lint/format/check rules or setup we should have.
- Updates CONTRIBUTING.md as necessary for the current state of things.

deboer-tim:
Charlie's #248 fixed most of this. I've gone through all the other differences in things like eslint and there's nothing that I think we should change short term, just confirming PR checks.

deboer-tim:
Lots of minor cleanup, enough for now.

PR #273, opened 2024-04-04
chore: strict tsconfig

deboer-tim:
### What does this PR do?

Following what AI Lab did, just upping our backend tsconfig settings.

### Screenshot / video of UI

N/A

Part of #247.

### How to test this PR?

Just adds additional checks to PR build.

### DIFF
--- a/packages/backend/src/build-disk-image.ts
+++ b/packages/backend/src/build-disk-image.ts
@@ -44,7 +44,7 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
 
   // VALIDATION CHECKS
   for (const { field, message } of requiredFields) {
-    if (!build[field]) {
+    if (!build[field as keyof BootcBuildInfo]) {
       await extensionApi.window.showErrorMessage(message);
       throw new Error(message);
     }
--- a/packages/backend/tsconfig.json
+++ b/packages/backend/tsconfig.json
@@ -1,10 +1,6 @@
 {
   "compilerOptions": {
-    "strictNullChecks": true,
-    "strictBindCallApply": true,
-    "alwaysStrict": true,
-    "strictFunctionTypes": true,
-    "noImplicitThis": true,
+    "strict":true,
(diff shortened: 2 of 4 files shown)


packages/backend/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
   "compilerOptions": {
     "strictNullChecks": true,
+    "strictBindCallApply": true,
+    "alwaysStrict": true,
+    "strictFunctionTypes": true,
+    "noImplicitThis": true,
```
benoitf, line 7: if possible go fo `"strict": true,`

deboer-tim: Done.

benoitf (approved):

packages/backend/src/build-disk-image.ts
```diff
@@ -44,7 +44,7 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History): P
 
   // VALIDATION CHECKS
   for (const { field, message } of requiredFields) {
-    if (!build[field]) {
+    if (!build[field as keyof BootcBuildInfo]) {
```
benoitf, line 47: for this one it's look tricky

deboer-tim: Yeah. I like that it actually says what it's doing though, and other options I tried were more verbose, e.g.:
`if (!(build as {[key: string]: any})[field]) {`
or:
`if (!(build as Record<string, any>)[field]) {`
or:
```
const obj: Record<string, any> = build;
if (!obj[field]) {
```

Prefer any of these or a different one?

deboer-tim: Merging, feel free to ping if you'd like me to change it :).

benoitf: ah yes no pb, I was hoping that we don't need to use any cast there but maybe it's not possible but anyway it's was not preventing any merge 👍

Issue #179, opened 2024-03-07
Review of the telemetry events


slemeur:
### Is your enhancement related to a problem? Please describe

We'd like to review the telemetry events we are reporting to ensure we have proper information about how the users are using the extension and where they are eventually facing challenges

### Describe the solution you'd like

We'd like to review the telemetry events we are reporting to ensure we have proper information about how the users are using the extension and where they are eventually facing challenges

deboer-tim:
Notes from meeting this morning:
- [ ] Log event for clicking on pull in the welcome
- [ ] Log event when switching to the build page
- [ ] Chart to show type of images being built
- [ ] Chart to show architecture of images being built

deboer-tim:
Telemetry in, charts updated, but forgot to list the implied charts required for the new telemetry. Will add these, then close.

cdrage:
https://github.com/containers/podman-desktop-extension-bootc/pull/288 is closed, I believe we can close this now?

deboer-tim:
I've verified all events are being output and appear in Segment, and have drafts of all charts. Closing.

PR #288, opened 2024-04-05
feat: add telemetry to pull, goto build page

deboer-tim:
### What does this PR do?

Exposed the telemetry logger from the extension, and changed build-disk-image.ts to use it instead of its own. While doing this I noticed that logUsage was being called before and after the build, so I removed the first to avoid double-counting.

Adds telemetry event when pulling images. The intent here is to show how often users pull the example but will also trigger when pulling the image builder itself, which might also be useful.

Exposes telemetry logging to frontend, and sends an event when someone clicks on the button to go to the Build page.

### Screenshot / video of UI

N/A

Part of #179.

### How to test this PR?

Tests added to confirm telemetry events. To test manually, enable telemetry and either watch live in Segment or modify PD telemetry to output to console.

### DIFF
--- a/packages/shared/src/BootcAPI.ts
+++ b/packages/shared/src/BootcAPI.ts
@@ -29,4 +29,5 @@ export abstract class BootcApi {
   abstract openFolder(folder: string): Promise<boolean>;
   abstract generateUniqueBuildID(name: string): Promise<string>;
   abstract openLink(link: string): Promise<void>;
+  abstract logUsage(eventName: string, data?: Record<string, unknown> | undefined): Promise<void>;
 }
--- a/packages/shared/src/BootcAPI.ts
+++ b/packages/shared/src/BootcAPI.ts
@@ -29,5 +29,6 @@ export abstract class BootcApi {
   abstract openFolder(folder: string): Promise<boolean>;
   abstract generateUniqueBuildID(name: string): Promise<string>;
   abstract openLink(link: string): Promise<void>;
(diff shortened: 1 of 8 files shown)


packages/shared/src/BootcAPI.ts
```diff
@@ -29,4 +29,5 @@ export abstract class BootcApi {
   abstract openFolder(folder: string): Promise<boolean>;
   abstract generateUniqueBuildID(name: string): Promise<string>;
   abstract openLink(link: string): Promise<void>;
+  abstract logUsage(eventName: string, data?: Record<string, unknown> | undefined): Promise<void>;
```
cdrage, line 32: to differentiate it from the other commands, we should name it telemetryLogUsage https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/shared/src/StudioAPI.ts#L57C12-L57C29

cdrage (changes requested):
Should we add logError as well? https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/shared/src/StudioAPI.ts#L58

packages/shared/src/BootcAPI.ts
deboer-tim: Renamed. As per other comment I added telemetryLogError as well, I just hadn't added that since we aren't using it yet.

Ready for another review.

cdrage (approved):
LGTM!

Issue #262, opened 2024-04-02
Requirements should be visible before building


deboer-tim:
### Is your enhancement related to a problem? Please describe

Our existing check to make sure the Podman machine is rootful (and by default our coming checks for Podman 5.0 and machine memory) don't get used until you try to run a build.

### Describe the solution you'd like

It would be more useful if these showed up before you try to build, e.g. on the homepage or build page. If they were here we could do warnings too, instead of just failing a build.

Issue #200, opened 2024-03-19
Clicking "No" on do you want to override, continues with build and times out.


cdrage:
### Bug description

TO ADD VIDEO: If you click "No" to not overriding a folder when building within the build page, it will continue the build and timeout. 

### Operating system

mac m1

### Version

next (development version)

PR #309, opened 2024-04-08
chore: expose validation API

deboer-tim:
### What does this PR do?

Currently, there is no way for the frontend to check prereqs or see if there is an existing image before attempting to build, at which point the backend has to open dialogs to notify of an error, or ask the user whether to overwrite.

This PR moves the existing code from the build to check prereqs and for existing images out into two functions accessible to the frontend:
 - checkPrereq() checks that the machine is Podman v5 and rootful.
 - buildExists() checks if there's an existing build in a given folder.

It also adds an option to the build to overwrite an existing image.

Tests existed for the underlying function and the build just calls the refactored functions, so there is no behaviour change from before.

### Screenshot / video of UI

N/A

Refactoring in preparation for #200 and #262.

### How to test this PR?

Existing PR checks and Build page works as before.

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -19,12 +19,13 @@
 import * as podmanDesktopApi from '@podman-desktop/api';
 import type { ImageInfo } from '@podman-desktop/api';
 import type { BootcApi } from '/@shared/src/BootcAPI';
-import type { BootcBuildInfo } from '/@shared/src/models/bootc';
+import type { BootcBuildInfo, BuildType } from '/@shared/src/models/bootc';
 import { buildDiskImage } from './build-disk-image';
 import { History } from './history';
 import * as containerUtils from './container-utils';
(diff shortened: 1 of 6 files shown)


benoitf (commented):
LGTM but it's missing tests

benoitf:
I saw Tests existed for the underlying function and the build just calls the refactored functions, so there is no behaviour change from before.

but as there are public methods, they should have individual tests

benoitf (approved):
approving expecting individual tests will come

cdrage (approved):
LGTM. Confirmed that this still works building the image no problem. 

However, from the function, it will still obviously prompt from within the API.

I'm assuming that next PR / will move the prompt elsewhere?

deboer-tim:
Added tests for both new functions.

> I'm assuming that next PR / will move the prompt elsewhere?

Yes, next PR will be for Build.svelte.

benoitf (approved):
approving again

PR #322, opened 2024-04-11
feat: build form validation

### LINKED ISSUE #200: Clicking "No" on do you want to override, continues with build and times out.

### LINKED ISSUE #262: Requirements should be visible before building

deboer-tim:
### What does this PR do?

Prior to this change, if you were missing prereqs (v5+rootful) or were missing some option you wouldn't see a problem until you build, and then after you accepted the error message it would sit on the build screen for 10s before it timed out. Likewise, if there was an existing build and you said 'no' to overwriting, it would wait for 10s to time out.

With this change, all validation is done inside the form. If you're missing a prereq or one of the options, the error message is visible at the bottom of the form and the Build button is disabled until you fix it. If you need to confirm overwriting an existing build, a checkbox appears.

Renamed buildErrorMessage so it's more clear this is the post-build error.

Added tests for prereqs and overwriting, and fixed the rather obvious error in api-impl.ts.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-bootc/assets/19958075/761049ef-f13a-47a0-b7b0-848ccc033ac7

Fixes #262. Fixes #200.

### How to test this PR?

Automated tests added. For manual testing, try without one of the prereqs (e.g. podman 4.9 or non-rootful, or with missing info in the form.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -139,7 +139,7 @@ async function buildBootcImage() {
     // the reason being is that the validation / error logic happens in buildDiskImage
     // in the backend and it will error out there as that is where we can console.log
     // as well as notify the user of the error via showErrorMessage / showInformationMessage, etc.
-    bootcClient.buildImage(buildOptions, overwrite || true); // TODO
+    bootcClient.buildImage(buildOptions, overwrite);
 
     // Continue doing listHistoryInfo until the build container name, tag, type and arch show up
     // this means we can safely exit and see it in the dashboard as it's now in the history / running in the background.
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -17,21 +17,20 @@ export let imageTag: string | undefined = undefined;
 
 // Image variables
 let selectedImage: string | undefined;
-let buildImageName: string;
-let buildTag: string;
-let buildEngineId: string;
+let existingBuild: boolean = false;
 
 // Build options
 let buildFolder: string;
 let buildType: BuildType[];
 let buildArch: string | undefined;
+let overwrite: boolean = false;
 
 // Other variable
 let success = false;
 let buildInProgress = false;
 let bootcAvailableImages: ImageInfo[] = [];
-let errorMessage = '';
-let errorFormValidation = '';
+let buildErrorMessage = '';
+let errorFormValidation: string | undefined = undefined;
 
 function findImage(repoTag: string): ImageInfo | undefined {
   return bootcAvailableImages.find(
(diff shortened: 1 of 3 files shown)


packages/frontend/src/Build.svelte
```diff
@@ -67,53 +66,80 @@ async function fillBuildOptions() {
...
   };
+
   buildInProgress = true;
   try {
     // Do not await.. just start the build.
     // the reason being is that the validation / error logic happens in buildDiskImage
     // in the backend and it will error out there as that is where we can console.log
     // as well as notify the user of the error via showErrorMessage / showInformationMessage, etc.
-    bootcClient.buildImage(buildOptions);
+    bootcClient.buildImage(buildOptions, overwrite || true); // TODO
```
cdrage, line 142: remove todo? it's always true?

packages/frontend/src/Build.svelte
```diff
@@ -67,53 +66,80 @@ async function fillBuildOptions() {
...
   const buildID = await bootcClient.generateUniqueBuildID(buildImageName);
 
   // The build options
+  const image = findImage(selectedImage);
   const buildOptions: BootcBuildInfo = {
     id: buildID,
     image: buildImageName,
-    tag: buildTag,
-    engineId: buildEngineId,
+    tag: selectedImage.split(':')[1],
```
cdrage, line 129: ?? unsure what's happening here with regards to splitting. what if the image has no associated tag / this is still okay?

packages/frontend/src/Build.svelte
```diff
@@ -67,53 +66,80 @@ async function fillBuildOptions() {
...
+    engineId: image?.engineId ?? '',
```
cdrage, line 130: don't we always have to have an engine id for the bootc iamge build info?

cdrage (changes requested):

packages/frontend/src/Build.svelte
deboer-tim: Doh, forget to remove this at the end. Fixed in new commit.

packages/frontend/src/Build.svelte
deboer-tim: Same as above, this is what it was doing before. I had to add the ? and ?? to make the compiler happy and provide a fallback, but there aren't any cases where the image wouldn't be found.

packages/frontend/src/Build.svelte
deboer-tim: This is actually what was there before [*], it was assigning the buildTag variable with the code below and then set here, and I removed the intermediate assigning to a variable. I haven't noticed any problems with it.

```
const image = findImage(selectedImage);
    if (image) {
      ...
      buildTag = selectedImage.split(':')[1];
      buildEngineId = image.engineId;
    }
```

[*] nit: if the image can't be found, in the existing code the tag and engine id would never be set, or worse would be set to the last image tag and engine id you selected that it could find. So this new code is 'better', except that I don't think there are any cases where it can't find the image.

cdrage (approved):

PR #351, opened 2024-04-18
feature: automatically select image based on image architecture inspect

cdrage:
feature: automatically select image based on image architecture inspect

### What does this PR do?

* Automatically selects the image based upon what architecture is
  available under inspect image.

Note that this is also in anticipation of manifest feature being added,
so there is some code in here (such as an array, rather than a single
string) for manifest.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/2d09c5c5-6164-4811-89c0-e17e813608ae

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/245

### How to test this PR?

Open a build, try and click on arm64 / amd64.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -19,6 +19,14 @@ export let imageTag: string | undefined = undefined;
 let selectedImage: string | undefined;
 let existingBuild: boolean = false;
 
+// Architecture variable
+// This is an array as we will be support manifests
+// the array contains the architectures (ex. arm64, amd64, etc.) of the image
+// we will ONLY enable the architectures that are available in the image / manifest within the form.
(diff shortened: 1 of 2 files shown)


packages/frontend/src/Build.svelte
```diff
@@ -66,6 +73,24 @@ async function fillBuildOptions() {
   }
 }
 
+async function fillArchitectures(historyInfo: BootcBuildInfo[]) {
```
deboer-tim, line 76: There's a lot of overlap with fillBuildOptions, which even sets buildArch in one case. It's fine as-is, but I think they should probably be merged.

deboer-tim (approved):

Issue #324, opened 2024-04-11
Multi-arch support: Detect manifest, extract which image to be used and then build with bib


cdrage:
### Is your feature request related to a problem? Please describe

We are trying to implement manifests into this extension.

The caveat is that we cannot directly pass in manifests to bootc-image-builder. This will **not** be supported due to upstream issues with the go library.

We should support being able to build from a manifest and use the correct image.

### Describe the solution you'd like

In order to do this we will:

1. Create the manifest via PD / CLI
2. List the images within this extension
3. If manifest, inspect the manifest
4. Get the digest id that matches the architecture selected
5. List images information and get the image id that matches the digest id
6. (TEMPORARY WORKAROUND UNTIL FIXED) bib does not support passing in a direct image id, we will have to copy / or rename the image to something like `foobar`
7. Pass in the image id / name and bib will build it successfully.

cdrage:
Catching up with @kingsleyzissou this will now be achievable within bootc-image-builder!

It has been tested in https://github.com/osbuild/images/pull/595 (see latest comments)

For the implementation, we will have to support listing manifests as well as using inspectManifest in order to achieve multi-platform selection.

cdrage:
Friday update: https://github.com/osbuild/bootc-image-builder/pull/397 is the last part to go in, once that is merged, we will have to wait for a few hours for konflux to push a new bootc-image-builder image and then we are good to go for updating the PR #356 for full support!

Issue #340, opened 2024-04-16
Add inspectManifest endpoint for bootc


cdrage:
### Is your enhancement related to a problem? Please describe

Required for https://github.com/containers/podman-desktop-extension-bootc/issues/324

### Describe the solution you'd like

Add the PD inspectManifest API endpoint to bootc so we can inspect manifests.

PR #356, opened 2024-04-19
feat: add manifest / multi-arch selection support

cdrage:
feat: add manifest / multi-arch selection support

### What does this PR do?

* Manifests are now listed, inspected and propagated within the Build
  page
* Simply select a manifest, select the arch and it will build

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-bootc/assets/6422176/5a2c32a2-cddc-4180-9758-c7512d48bbf8

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/324

### How to test this PR?

Follow the video above, or do the following tests:

1. Build a manifest within Podman Desktop (select two architectures, and
   build a bootc image). Must use the latest Podman Desktop
2. Select the manifest within the extension and see it build.

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -100,12 +100,30 @@ export class BootcApiImpl implements BootcApi {
   async listBootcImages(): Promise<ImageInfo[]> {
     let images: ImageInfo[] = [];
     try {
-      const retrieveImages = await podmanDesktopApi.containerEngine.listImages();
-      images = retrieveImages.filter(image => {
-        if (image.Labels) {
-          return image.Labels['bootc'] ?? image.Labels['containers.bootc'];
+      const retrievedImages = await podmanDesktopApi.containerEngine.listImages();
+      const filteredImages: ImageInfo[] = [];
+      for (const image of retrievedImages) {
+        let includeImage = false;
+
+        // The image must have RepoTags and Labels to be considered a bootc image
+        if (image.RepoTags && image.Labels) {
+          // Convert to boolean by checking the string is non-empty
+          includeImage = !!(image.Labels['bootc'] ?? image.Labels['containers.bootc']);
+        } else if (image?.isManifest) {
+          // Manifests **usually** do not have any labels. If this is the case, we must find the images associated to the
+          // manifest in order to determine if we are going to return the manifest or not.
+          const manifestImages = await containerUtils.getImagesFromManifest(image, retrievedImages);
+          // Checking if any associated image has a non-empty label
+          includeImage = manifestImages.some(
+            manifestImage => !!(manifestImage.Labels['bootc'] ?? manifestImage.Labels['containers.bootc']),
+          );
         }
-      });
+
+        if (includeImage) {
+          filteredImages.push(image);
+        }
+      }
+      images = filteredImages;
     } catch (err) {
       await podmanDesktopApi.window.showErrorMessage(`Error listing images: ${err}`);
       console.error('Error listing images: ', err);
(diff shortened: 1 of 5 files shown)


cdrage:
**IMPORTANT NOTES**

There are moving parts that need to be done before this merges:

- [ ] Added more tests (one is currently commented out)
- [ ] https://github.com/containers/podman-desktop-extension-bootc/issues/340 merged
- [ ] https://github.com/osbuild/images/pull/595 merged
- [ ] bootc-image-builder updated with latest osbuild/images changes
- [ ] update the image (it has been changed to a temporary one) in this PR to the updated bib image.

However, this is ready for review and testing (see changes in the last commit) 💯

packages/backend/src/container-utils.ts
```diff
@@ -317,3 +317,31 @@ export async function removeContainerAndVolumes(engineId: string, container: str
...
+// 5. Return the imageInfo of all matching images (these are the images that are part of the manifest)
+export async function getImagesFromManifest(image: extensionApi.ImageInfo): Promise<extensionApi.ImageInfo[]> {
+  if (!image.isManifest) {
+    throw new Error('Image is not a manifest');
+  }
+
+  // Get the manifest
+  const manifest = await inspectManifest(image.engineId, image.Id);
+
+  // Get the digest values, if there are no digests, make sure array is [] so we don't run into
```
deboer-tim, line 336: If no digests then doesn't this return an empty list? If so seems more straightforward to just use an if and return early.

packages/frontend/src/Build.spec.ts
```diff
@@ -80,6 +80,79 @@ const mockBootcImages: ImageInfo[] = [
   },
 ];
 
+const mockImageInspect: ImageInspectInfo = {
```
deboer-tim, line 83: There is so much boilerplate/default values here I am wondering if it is a case to remove most of it and use `as unknown as ImageInspectInfo` :)

packages/backend/src/api-impl.ts
```diff
@@ -101,11 +101,34 @@ export class BootcApiImpl implements BootcApi {
...
       const retrieveImages = await podmanDesktopApi.containerEngine.listImages();
-      images = retrieveImages.filter(image => {
+      const filteredImages: ImageInfo[] = [];
+      for (const image of retrieveImages) {
+        let includeImage = false;
+
+        // If the image has no RepoTags, it is likely a <none> / tmp image and we should not be
+        // displaying it in the list of images.
+        if (!image.RepoTags) {
+          continue;
```
deboer-tim, line 111: The flow here is a little odd to me where this one case uses `continue`, but others set `includeImage` (or not). It would be more consistent to do similar logic in all cases, i.e. either use continue to avoid all invalid cases and always do a push at the end, or this case just doesn't set includeImage.

cdrage:
Removed the WIP title and this is ready for a preliminary review.

However.. don't merge until https://github.com/osbuild/images/pull/595 is closed and completed, and this PR has an updated bib image.

cdrage:
https://github.com/osbuild/bootc-image-builder/pull/397 has been merged!

This PR is ready for review!

Please throughly test building a manifest from within Podman Desktop, and test it out with bootc (build both arm64 and amd64 please!)

deboer-tim (approved):
Awesome work Charlie!

I was able to build several manifests using Podman Desktop 1.10, and they were as seamless (just twice as long!) as building for one platform.

I've verified the Build page automatically makes the correct platform(s) available as you switch between various images of either platform, or a multi-arch manifest.

I've verified that using a manifest and picking the ARM option builds and creates a usable image. Picking the AMD option clearly builds a different disk image and takes longer. I haven't had a chance to try running the built image on an AMD machine, but since that's not directly part of this PR I'm happy to approve already.

PR #359, opened 2024-04-22
chore: send telemetry on startup

deboer-tim:
### What does this PR do?

I noticed that AI lab is doing this - send the extension version in on startup so that we can see which version users have installed. This should also be in the PD telemetry, but easier to track/correlate here.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Test updated.

### DIFF
--- a/packages/backend/src/extension.ts
+++ b/packages/backend/src/extension.ts
@@ -45,9 +45,7 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
     );
   }
 
-  telemetryLogger.logUsage('start', {
-    version: version,
-  });
+  telemetryLogger.logUsage('start');
 
   const history = new History(extensionContext.storagePath);
   await history.loadFile();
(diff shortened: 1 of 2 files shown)


packages/backend/src/extension.ts
```diff
@@ -45,6 +45,10 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
     );
   }
 
+  telemetryLogger.logUsage('start', {
+    version: version,
```
benoitf, line 49: AFAIK if you do start, podman desktop will enrich the data with the extension version automatically

benoitf: I mean, podman desktop add always extra properties including the version

deboer-tim: Ah, I was looking at all the other properties and didn't notice it was adding version there: properties/common.extensionVersion. Confirmed and PR changed.

I noticed load/activate extension events from PD only include the extension id and not version, might be a better long-term practice if we just added it there.

benoitf (approved):

cdrage (approved):

PR #364, opened 2024-04-23
chore(ci,tests): add workflow to run bootc e2e tests on windows

odockal:
### What does this PR do?
Add nightly/on demand workflow to run bootc e2e tests on windows
https://github.com/containers/podman-desktop-e2e/issues/172

### How to test this PR?
It would require adding scepcial credentials into repo. TBD

### DIFF
--- a/.github/workflows/bootc-e2e-nightly-windows.yaml
+++ b/.github/workflows/bootc-e2e-nightly-windows.yaml
@@ -1,3 +1,20 @@
+#
+# Copyright (C) 2024 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
 name: Podman Desktop BootC E2E Nightly
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/bootc-e2e-nightly-windows.yaml
```diff
@@ -0,0 +1,236 @@
+name: Podman Desktop BootC E2E Nightly
```
benoitf, line 1: FYI, missing the copyright headers

.github/workflows/bootc-e2e-nightly-windows.yaml
```diff
@@ -0,0 +1,236 @@
...
+      env:
+        DEFAULT_FORK: 'containers'
+        DEFAULT_BRANCH: 'main'
+        DEFAULT_EXT_TESTS: '1'
+        DEFAULT_NPM_TARGET: 'test:e2e'
+        DEFAULT_ENV_VARS: 'TEST_PODMAN_MACHINE=true'
+        DEFAULT_PODMAN_OPTIONS: 'INIT=1,START=1,ROOTFUL=1,NETWORKING=0'
+        DEFAULT_EXT_REPO_OPTIONS: 'REPO=podman-desktop-extension-bootc,FORK=containers,BRANCH=main'
+        DEFAULT_VERSION: '5.0.2'
+        DEFAULT_URL: 'https://github.com/containers/podman/releases/download/v5.0.2/podman-remote-release-windows_amd64.zip'
```
benoitf, line 85: you need to update all extensions when there is a new version of podman included in Podman Desktop ?

shouldn't be part of some data grabbed from podman desktop

benoitf (commented):
it looks like the dependency to podman should be externalized to this repository$

this repo should only depend on 'Podman Desktop' thing

.github/workflows/bootc-e2e-nightly-windows.yaml
odockal: That is a good idea.

.github/workflows/bootc-e2e-nightly-windows.yaml
odockal: done

odockal:
> it looks like the dependency to podman should be externalized to this repository$
> 
> this repo should only depend on 'Podman Desktop' thing

@benoitf After the review feedback I have changed the workflow to use the podman version used by podman-desktop when triggered on nightly base, and then there is an option to run workflow on demand with podman version to be choose by the user. But the param's value is static, I do think the default value can be extracted dynamically based on what is in podman-desktop. If that is what you meant?

benoitf (approved):

odockal:
Great, I was not sure first. Also it came to me, we could probably make it default behavior, so it would be incorporated in the pde2e images but still it would be great to have another option to override default podman version from outside. I have filled an issue for this: https://github.com/containers/podman-desktop-e2e/issues/177.

PR #388, opened 2024-04-25
docs: add port forwarding for testing

cdrage:
docs: add port forwarding for testing

### What does this PR do?

* Fixes up the order of the commands
* Adds notes / information about using port forwarding on the qemu
  commands

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Run the command after build an arm64 image, specifically the test image, and curl the 8080 port.

### DIFF
--- a/docs/vm_guide.md
+++ b/docs/vm_guide.md
@@ -24,25 +24,31 @@ brew install qemu
 
 **Usage:**
 
+**Important note:** Assuming you are running the example application, the command will port forward local port 8080 to 80. This can be changed in the below `qemu` command.
+
 1. Build a RAW image
 2. Run the following command:
 ```sh
 # Change to your VM image location
 export DISK_IMAGE=/Users/myusername/bootc/image/disk.raw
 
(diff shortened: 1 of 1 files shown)


benoitf (approved):
don't know for the 'We' wording in `We assume you are` vs `Assuming you are`

cdrage:
> don't know for the 'We' wording in `We assume you are` vs `Assuming you are`

you're right, that's incorrect wording! thank you for pointing that out. updated the PR.

docs/vm_guide.md
```diff
@@ -91,20 +98,26 @@ brew install qemu
 
 **Usage:**
 
+**Important note:** Assuming you are running the example application, we will port forward local port 8080 to 80. This can be changed in the below `qemu` command.
```
benoitf, line 101: `we will port forward` --> `the command will`

deboer-tim (approved):
Tested, works. E2E failure is known and unrelated. LGTM

Issue #382, opened 2024-04-24
Update readme for extension catalog


deboer-tim:
### Is your enhancement related to a problem? Please describe

The bootc extension will now appear in the Podman Desktop extension catalog.

### Describe the solution you'd like

We need to:
- Update the installation instructions
- Make sure the readme looks good inside PD (e.g. likely means absolute paths for images)

Close relationship with and can be done at the same time as #338.

Issue #338, opened 2024-04-16
Use gifs not videos in README for it to look good in Podman Desktop


cdrage:
### Is your enhancement related to a problem? Please describe

With https://github.com/containers/podman-desktop/pull/6775 being added, we cannot use videos.

Use gifs instead to display the information in the readme.

### Describe the solution you'd like

FYI I've been using this to convert the mp4's in the directory into a reasonably sped up gif:

```
for f in *.mp4; do ffmpeg -i "$f" -vf "fps=10" "%04d.png" && gifski --fps 10 -o "${f%.mp4}.gif" --width 800 *.png && rm *.png; done
```

### Additional context

We should do this before 1.0.0 release

PR #411, opened 2024-04-29
docs: update readme with simplified information and added gifs

cdrage:
docs: update readme with simplified information and added gifs

### What does this PR do?

* Updates the readme steps to be a lot more simplified (we no longer
  need to push.. the logs are shown in the folder / can be found easily,
  etc.)
* Added gifs

### Screenshot / video of UI

See README.md

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/382
Closes https://github.com/containers/podman-desktop-extension-bootc/issues/338

### How to test this PR?

View the README.md, make sure that the gifs are there. Make sure that
the steps are accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,55 +131,46 @@ podman machine start
 
 ## Installation
 
-This extension can be installed through the **Extensions** page of Podman Desktop:
+This extension can be installed through the **Extensions** section of Podman Desktop:
 
-1. Go to *Settings > Extensions*.
-2. Copy and paste the below image to the *Name of the image* field:
+1. Go to *Extensions* in the navbar.
+2. Copy and paste the below image to the *Install custom...* section:
 
 ```
 ghcr.io/containers/podman-desktop-extension-bootc
 ```
 
-3. Click *Install extension from the OCI image*.
+3. Click *Install*.
 
-[![](/docs/img/install_extension.gif)](https://github.com/containers/podman-desktop-media/raw/bootc-extension/videos/install_extension.mp4)
+![](https://raw.githubusercontent.com/containers/podman-desktop-media/bootc-extension/gifs/install_extension_new.gif)
 
 ## Usage
 
 1. **Build your bootc-enabled Containerfile:**
 
-> In our example, we are going to change the root password for testing purposes when accessing the OS.
+> In the example below, we are going to change the root password for testing purposes when accessing the OS. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,55 +125,46 @@ podman machine start
 
 ## Installation
 
-This extension can be installed through the **Extensions** page of Podman Desktop:
+This extension can be installed through the **Extensions** section of Podman Desktop:
 
-1. Go to *Settings > Extensions*.
-2. Copy and paste the below image to the *Name of the image* field:
+1. Go to *Extensions* in the navbar.
+2. Copy and paste the below image to the *Install custom...* section:
```
deboer-tim, line 131: We should direct users to the Catalog instead. By early next week we would be listed there, and easier than copy/paste.

README.md
```diff
@@ -125,55 +125,46 @@ podman machine start
...
 
-[![](/docs/img/install_extension.gif)](https://github.com/containers/podman-desktop-media/raw/bootc-extension/videos/install_extension.mp4)
+![](https://raw.githubusercontent.com/containers/podman-desktop-media/bootc-extension/gifs/install_extension_new.gif)
 
 ## Usage
 
 1. **Build your bootc-enabled Containerfile:**
 
-> In our example, we are going to change the root password for testing purposes when accessing the OS.
+> In the below example, we are going to change the root password for testing purposes when accessing the OS.
```
deboer-tim, line 145: `below example` -> `example below`? Just reads easier to me.

deboer-tim (changes requested):

README.md
cdrage: I agree, should we update this once it's in the catalog?

deboer-tim: We have a chicken and egg issue here: if we commit a change now then it'll be wrong for anyone who tries to install over the next week; if we don't then the copy in the catalog would contain text on installing via custom link. Unless we patch the catalog next week to not match what was in the release, which is also kind of ugly...

Since there will be far more people looking at this next week, I think the simplest is to just update it now. If anyone can't figure out how to install in the meantime hopefully they'll ping us.

cdrage: added https://github.com/containers/podman-desktop-extension-bootc/issues/413

deboer-tim (approved):

PR #449, opened 2024-05-06
feat: support rhel bib

### LINKED ISSUE #428: Support for RHEL image builder

Support for different/arbitrary builders is tracked via #238 and isn't feasible right now: there is no API/stability and doing so now would give the user too many ways to shoot themselves in the foot. However, now that registry.redhat.io/rhel9/bootc-image-builder is released we will have some users that want to use it and we could do this in a more controlled manner.

Test registry.redhat.io/rhel9/bootc-image-builder:9.4 to see if it is compatible/stable enough, or find a version that is equivalent to the current centos builder. Check containers for the bootc.diskimage-builder label, and if it exists and points to the rhel builder, then switch to and use the specific version we've tested.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds a new preference that allows you to switch between:
- 'image' - the default, picks up builder from image label (or use centos if none is defined)
- 'centos' - always use Centos builder
- 'rhel' - always use RHEL builder

createBuilderImageOptions() is made async to allow it to check the image and switch the builder if necessary.

### Screenshot / video of UI

N/A, only new preference visible.

Fixes #428.

### How to test this PR?

Tests added for the change, but need to test RHEL bib in general.


packages/backend/src/build-disk-image.ts
```diff
@@ -305,8 +305,27 @@ export async function getUnusedName(name: string): Promise<string> {
...
+
+  // default to centos builder
+  let builder = bootcImageBuilderCentos;
+  if (buildProp === 'rhel') {
+    // switch to rhel if that's the preference
+    builder = bootcImageBuilderRHEL;
+  } else if (buildProp === 'image') {
+    // or switch to rhel if the preference comes from the image label
+    // AND we detect the rhel label
+    const buildLabel = await containerUtils.getImageBuilderLabel(image);
```
cdrage, line 322: listImages is very expensive call and I see it's done within getImageBuilderLabel and each time we do it it makes the build "stutter" a bit before building.

i think that we should move this earlier in the call process and use the already done listImages call in the code? so we can avoid calling it twice.

packages/backend/src/build-disk-image.ts
```diff
@@ -305,8 +305,27 @@ export async function getUnusedName(name: string): Promise<string> {
...
+    if (buildLabel === 'registry.redhat.io/rhel9/bootc-image-builder') {
+      builder = bootcImageBuilderRHEL;
```
cdrage, line 324: this part is confusing to me for the if statement.

If buildProp = rhel, we switch to rhel

i feel like we dont need this check statement as it'll just be a:

if rhel = use this image

if centos or default, use centos

But that's it, no need to rhel check?

cdrage (changes requested):

packages/backend/src/build-disk-image.ts
deboer-tim: Where do we already call listImages()? I agree we should not call twice if we can avoid it but I didn't see an obvious use already.

packages/backend/src/build-disk-image.ts
deboer-tim: See https://github.com/CentOS/centos-bootc/pull/453 for some background. The `bootc.diskimage-builder` label allows an image to say 'I expect to be built with builder X'.

If the default preference (image) is used, then we need to look for this label. For now, we only look for one specific value and change to using the RHEL builder. So there are two cases where we'd use the RHEL builder: the preference is explicitly set to prefer RHEL, or preference is image and we detected an image that has asked for this builder.

We could remove the 'image' option and the image label detection for now if you're worried about the cost; users could still switch the builder (for all builds) through the preference.

cdrage: for now I think that we should just have it 100% through preferences.

it's also because I just inspected a few of my images and there is no `bootc.diskimage-builder` for the fedora bootc image...

ex:

```
    "Labels": {
      "containers.bootc": "1",
      "io.buildah.version": "1.35.3",
      "org.opencontainers.image.version": "40.20240503.0",
      "ostree.bootable": "true",
      "ostree.commit": "0b148652f58f77457b63129ae2f95d3f742b521b358727e2daca81fe7cd49e25",
      "ostree.final-diffid": "sha256:12787d84fa137cd5649a9005efe98ec9d05ea46245fdc50aecb7dd007f2035b1",
      "ostree.linux": "6.8.8-300.fc40.x86_64",
      "rpmostree.inputhash": "2d27d73eaa1cf6595b3b5e555ffe0b36a6854740298fa498e8bc1e4872f1d401"
    }
  },
  ```

Comment (deboer-tim) [reply packages/backend/src/build-disk-image.ts]: Ok, label detection code removed and simplified just to preferences for now.

File: packages/backend/package.json [hunk 3]
```diff
@@ -20,6 +20,15 @@
           "default": 30,
           "maximum": 120,
           "description": "Build timeout (in minutes)"
+        },
+        "bootc.builder": {
+          "type": "string",
+          "default": "Centos",
```
cdrage, line 26: should be CentOS throughout code + options

cdrage (changes requested):

packages/backend/src/build-disk-image.ts
```diff
@@ -304,8 +307,24 @@ export async function getUnusedName(name: string): Promise<string> {
...
+export async function getBuilder(): Promise<string> {
+  // use the preference to decide which builder to use
+  const buildProp = await getConfigurationValue<string>('builder');
+
+  if (buildProp === 'RHEL') {
+    return bootcImageBuilderRHEL;
+  }
+
+  // always default to centos bib
+  return bootcImageBuilderCentos;
```
cdrage, line 319: should we fail at all if the config setting is not centos or rhel? i see in our tests it still passed.

example you can pass in anything 'foobar' as the configuration setting and it'll always pick centos.

cdrage (changes requested):
should we add a test if configuration setting is blank? should pick centos (i'm assuming).

packages/backend/package.json
deboer-tim: Fixed.

packages/backend/src/build-disk-image.spec.ts
```diff
@@ -230,3 +253,29 @@ test('check build exists', async () => {
...
+    type: ['ami'],
+  } as BootcBuildInfo;
+  const builder = await getBuilder(build);
+
+  expect(builder).toBeDefined();
+  expect(builder).toEqual(bootcImageBuilderRHEL);
+});
+
+test('check uses Centos builder', async () => {
+  configurationGetConfigurationMock.mockReturnValue('centos');
```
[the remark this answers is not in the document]
deboer-tim: Fixed.

packages/backend/src/build-disk-image.ts
deboer-tim: I wanted to play it safe, i.e. even if config corrupted => we ignore and use CentOS anyway.

packages/backend/src/constants.ts
```diff
@@ -17,5 +17,6 @@
  ***********************************************************************/
 
 // Image related
-export const bootcImageBuilderContainerName = '-bootc-image-builder';
-export const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest-1714633180';
+export const bootcImageBuilder = 'bootc-image-builder';
+export const bootcImageBuilderCentos = 'quay.io/centos-bootc/bootc-image-builder:latest-1714474808';
```
cdrage, line 21: you changed the image to `latest-1714474808` which is untested

cdrage (changes requested):

packages/backend/src/constants.ts
deboer-tim: Doh, forgot I was testing that on the side. Fixed.

cdrage (approved):
Approved and tested! Tested via centos, fedora and rhel and switched between bib images.

HOWEVER there is one bug.

If you select ext4 or xfs with rhel bib selected, it will not build. This is because the rhel bib image does not have `--rootfs` yet.

We should be aware of this in case users run into that edge case, but the bib image for rhel will be updated shortly.

packages/backend/src/build-disk-image.ts
```diff
@@ -304,8 +307,24 @@ export async function getUnusedName(name: string): Promise<string> {
   return unusedName;
 }
 
+export async function getBuilder(): Promise<string> {
```
cgwalters, line 310: Hmm the commit message mentions the `bootc.diskimage-builder` label but I don't see it mentioned or used in the code anywhere?

Oh except...argh, we may have lost that label in some reworking of the base image builds.

packages/backend/src/constants.ts
```diff
@@ -17,5 +17,6 @@
  ***********************************************************************/
 
 // Image related
-export const bootcImageBuilderContainerName = '-bootc-image-builder';
-export const bootcImageBuilderName = 'quay.io/centos-bootc/bootc-image-builder:latest-1714633180';
+export const bootcImageBuilder = 'bootc-image-builder';
+export const bootcImageBuilderCentos = 'quay.io/centos-bootc/bootc-image-builder:latest-1714633180';
+export const bootcImageBuilderRHEL = 'registry.redhat.io/rhel9/bootc-image-builder:9.4';
```
cgwalters, line 22: There's more than one RHEL version. The 9.4 image is Tech Preview even and isn't going to be updated more and we don't want to use it by default for future RHEL versions like 9.6 or beyond.

I think in general we can just use `registry.redhat.io/rhel$MAJOR/bootc-image-builder:latest` reasonably safely.

To determine `$MAJOR` though requires inspecting the container image; like `podman run <image> bash -c '. /usr/lib/os-release && echo $REDHAT_SUPPORT_PRODUCT_VERSION` or so perhaps.

(In theory we could use `$VERSION_ID` except that may be overridden by downstream builds)

packages/backend/src/build-disk-image.ts
deboer-tim: We didn't see it used consistently enough in the images (and sounds like it isn't going to be?), so the extension still uses a simple builder preference (under Settings > Preferences > Bootable Containers in the UI).

PR #508, opened 2024-05-28
README.md: Trim links a bit

cgwalters:
The "source" link for centos-bootc changed; just drop it, as I'd like to have the documentation be canonical.

(Now of course hopefully we aren't going to go through a round
 of moving things *again* any time in the near future, but still
 I think users should click through to the docs site as
 a general rule)


cdrage:
Are we able to keep the examples link? I find it very useful to see what other maintained examples are / could reference against. 

I've seen other communities (ublue) reference them which was good affirmation.

Same for the "FROM part". It helps when copying and pasting to the Containerfile.

cgwalters:
OK, I reworked this to just drop the source links.

cdrage:
LGTM! I agree on the source link. That shouldn't be shown. Thank you!

PR #513, opened 2024-05-30
chore(tests): remove unused param

cbr7:
### What does this PR do?
Removes an unused param from a test method.

https://github.com/containers/podman-desktop-extension-bootc/issues/514

### DIFF
--- a/tests/playwright/src/bootc-extension.spec.ts
+++ b/tests/playwright/src/bootc-extension.spec.ts
@@ -62,8 +62,8 @@ beforeAll(async () => {
   navBar = new NavigationBar(page);
   const dashboardPage = await navBar.openDashboard();
   await playExpect(dashboardPage.heading).toBeVisible();
-  await playExpect(dashboardPage.podmanStatusLabel).toBeVisible({ timeout: 30000 });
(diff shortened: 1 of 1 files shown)


benoitf (approved):

tests/playwright/src/bootc-extension.spec.ts
```diff
@@ -61,6 +60,10 @@ beforeAll(async () => {
   const welcomePage = new WelcomePage(page);
   await welcomePage.handleWelcomePage(true);
   navBar = new NavigationBar(page);
+  const dashboardPage = await navBar.openDashboard();
+  await playExpect(dashboardPage.heading).toBeVisible();
+  await playExpect(dashboardPage.podmanStatusLabel).toBeVisible({ timeout: 30000 });
```
odockal, line 65: I would lower the timeout here, so in total we do not exceed the basic hook timeout. also, podman extension should get up quite fast (<10s)

odockal (approved):

tests/playwright/src/bootc-extension.spec.ts
cbr7: lowered to 10s

odockal:
@cbr7 Looks good, if the e2e tests are alright, we can leave the draft and have it merged.

Issue #487, opened 2024-05-22
Support pushing AMI to S3 bucket


vrothberg:
### Is your feature request related to a problem? Please describe

Copied from https://issues.redhat.com/browse/RHDP-980

### Describe the solution you'd like

BIB has AMI specific flags for becoming aware of your AWS creds & S3 bucket, so it will push the AMI to S3 and register it so it can then be used to launch an instance. Without these flags, the following has to happen outside of the bootc extension:

    The AMI file will be created as /path/provided/in/desktop/gui/ami/disk.raw
    Must push to AWS s3 bucket, you need to use the aws CLI like so:

        aws s3 cp /path/to/disk.raw s3://your-s3-bucket-name aws ec2 import-snapshot --description "rhel94-bootc" --disk-container [file://ami-container.json](file:///) 
    Must visit the AWS cloud console to create an AMI from the snapshot and wait for it to register before launching an instance.

Without the AWS flags exposed in podman-desktop bootc extension, I would 100% of the time use the terminal to build an AMI with BIB. If we will feature building an AMI during any demo of the bootc extension, these flags should be exposed. 
Actionable item == expose these flags in bootc extension when building an AMI:

--aws-ami-name
--aws-bucket
--aws-region

PR #576, opened 2024-06-25
chore: adopt checkbox component

deboer-tim:
### What does this PR do?

The temporary checkbox had some minor UI issues, time to use the Checkbox component from Podman Desktop UI library. There is a limitation that you cannot bind to a group array with a value that would have made this particular case a little simpler, but I just used a function that updates the buildType array.

### Screenshot / video of UI

Before:

Screenshot 2024-06-25 at 1 30 44 PM

After:

Screenshot 2024-06-25 at 1 33 46 PM

### What issues does this PR fix or reference?

Fixes #537 (and also required for light mode #540).

### How to test this PR?

Use the build form.

Issue #540, opened 2024-06-12
Light mode for form


cdrage:
### Is your enhancement related to a problem? Please describe

Added light mode for form

### Describe the solution you'd like

Light mode

deboer-tim:
I'm going to mark this done. All light mode variables are being picked up from Podman Desktop correctly, and fonts/size match the rest of the UI. The biggest remaining issue is the select dropdown at the top of the form, but that is an issue that needs to be solved in Podman Desktop and then we'll pick up the component here later.

PR #579, opened 2024-06-25
feat: add AWS uploading image

cdrage:
feat: add AWS uploading image

### What does this PR do?

* Adds the ability to upload to AWS through build
* Toggled advanced features section

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/487

### How to test this PR?

1. Add your credentials to ~/.aws/credentials
2. Build your image + provide your bucket, aminame, etc.


cdrage:
Encountered https://github.com/osbuild/bootc-image-builder/issues/501 when implementing this as I cannot self-test this with my minio server.

Tasks:
- [X] Works with AWS S3
- [ ] Add unit tests.

cdrage:
Ready for review.. although bit of a chicken and an egg problem. We can get this in, but we would need to rebase and update against #540 , or we wait for #540 and then rebase against that.

Totally fine either way, but there **are** some hardcoded color files in this PR since they do not match the other current ones when using from color-registry.ts

cc @deboer-tim

deboer-tim:
> Ready for review.. although bit of a chicken and an egg problem. We can get this in, but we would need to rebase and update against #540 , or we wait for #540 and then rebase against that.

Did you mean rebasing on #576? (as we don't have any other PRs up in this area that would require rebasing and there are several steps to do #540). I don't have an issue with this from a light mode standpoint, as we could easily replace these color values at the same time as fixing the others and it doesn't make sense to do anything else in the meantime.

cdrage:
> > Ready for review.. although bit of a chicken and an egg problem. We can get this in, but we would need to rebase and update against #540 , or we wait for #540 and then rebase against that.
> 
> Did you mean rebasing on #576? (as we don't have any other PRs up in this area that would require rebasing and there are several steps to do #540). I don't have an issue with this from a light mode standpoint, as we could easily replace these color values at the same time as fixing the others and it doesn't make sense to do anything else in the meantime.

Ah I mean if we should wait for light mode form to go in first before this PR. If you are good with hardcoded colours (for now), I'm fine with that!

👍

deboer-tim (approved):

PR #604, opened 2024-07-08
chore: light mode, mostly for build page

deboer-tim:
### What does this PR do?

Move up to latest ui-svelte in order to do the following. Requires a recent Podman Desktop build in order for variables to be defined at runtime.

In the Build form, pick up status colors, switch to use ErrorMessage component, and pick up other light mode color variables. Uses the same styling as PD for the tiles, and radio buttons use the checkbox color variables (since PD has no vars for radio inputs).

Use status colors in bootc status icon for now - we should probably reuse StatusIcon now, but that's separate work.

After doing this, we can even set the default text color in the index.

### Screenshot / video of UI

Screenshot 2024-07-08 at 12 16 46 PM

Not marking this as 'fixes' since we still need a select component, but this is most of #540.

### How to test this PR?

Check the homepage and build page in light and dark modes.

### DIFF
--- a/packages/frontend/index.html
+++ b/packages/frontend/index.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0" />
     <title>BootC Extension</title>
   </head>
-  <body class="overflow-auto text-white">
+  <body class="overflow-auto text-[var(--pd-default-text)]">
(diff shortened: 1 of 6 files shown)


cdrage (approved):
👍 built and looked at the form and great progress. Other issue is in progress regarding the select component.

LGTM

Issue #662, opened 2024-07-23
Running ISO build in e2e test takes too long


cbr7:
### Bug description

It seem creating an iso build on windows takes much longer than the other formats and it requires special handling and custom wait times.

### Operating system

Windows 11

### Version

next (development version)

PR #669, opened 2024-07-25
chore(test): special handling for iso builds

cbr7:
### What does this PR do?
Adds a special flag to build ISO images and increases timeout in such a case, because ISO builds take much longer than others.

https://github.com/containers/podman-desktop-extension-bootc/issues/662

### DIFF
--- a/tests/playwright/src/bootc-extension.spec.ts
+++ b/tests/playwright/src/bootc-extension.spec.ts
@@ -128,12 +128,11 @@ describe('BootC Extension', async () => {
 
       describe.skipIf(isLinux).each(['QCOW2', 'AMI', 'RAW', 'VMDK', 'ISO'])('Building images ', async type => {
         test(`Building bootable image type: ${type}`, async context => {
-          if(type === 'ISO'){
-            if(buildISOImage){
+          if (type === 'ISO') {
+            if (buildISOImage) {
               timeoutForBuild = 1200000;
               console.log(`Building ISO image requested, extending timeout to ${timeoutForBuild}`);
-            }
-            else{
+            } else {
               console.log(`Building ISO image not requested, skipping test`);
               context.skip();
             }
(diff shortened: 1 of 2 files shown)


tests/playwright/src/model/bootc-page.ts
```diff
@@ -20,6 +20,7 @@ import type { Locator, Page } from '@playwright/test';
 import { expect as playExpect } from '@playwright/test';
 import { waitUntil } from '@podman-desktop/tests-playwright';
 import { ArchitectureType } from '@podman-desktop/tests-playwright';
+import { time } from 'node:console';
```
odockal, line 23: unused import?

cbr7: yeah, fixed.

odockal (approved):

tests/playwright/src/bootc-extension.spec.ts
```diff
@@ -124,9 +126,18 @@ describe('BootC Extension', async () => {
         await playExpect.poll(async () => await imagesPage.waitForImageExists(imageName)).toBeTruthy();
       }, 150000);
 
-      test.skipIf(isLinux).each(['QCOW2', 'AMI', 'RAW', 'VMDK', 'ISO'])(
-        `Building bootable image type: %s`,
-        async type => {
+      describe.skipIf(isLinux).each(['QCOW2', 'AMI', 'RAW', 'VMDK', 'ISO'])('Building images ', async type => {
```
benoitf, line 129: I'm not sure if it's possible but shouldn't it be something like

```ts
const platforms = '['QCOW2', 'AMI', 'RAW', 'VMDK'];
if (!buildISOImage) {
   platforms.push('ISO')
}

 describe.skipIf(isLinux).each(platforms)('Building images ', async type => {
```

cbr7: Does that work? I was under the impression that vitest creates the test context before anything else, all the examples for parametrized tests that I've seen have the parameters provided directly to the each() without any previous logic.

benoitf: No idea it's why I emitted some doubts :-)

cbr7: @benoitf I just tried it, it seems that you can send params to the test just using a variable, however you can't manipulate the number of tests, basically it uses whatever value `platforms` has if defined globally or it reads it undefined and does not account for any changes to it after the test context is created.

So yes it seems that all the test contexts are created before any logic is executed.

benoitf: Ok thanks so ignore my comment

Issue #472, opened 2024-05-15
Support chown from bootc-image-builder


cdrage:
### Is your feature request related to a problem? Please describe

We're missing passing in the chown parameter: https://github.com/osbuild/bootc-image-builder?tab=readme-ov-file#detailed-description-of-optional-flags

### Describe the solution you'd like

Add the chown option

PR #695, opened 2024-07-31
feat: add chown (linux only)

cdrage:
feat: add chown (linux only)

### What does this PR do?

* Adds chown to the advanced settings page for build
* Only applicable to Linux (see:
  https://github.com/osbuild/bootc-image-builder/issues/576)

### Screenshot / video of UI

Screenshot 2024-07-31 at 2 42 38 PM

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/472

### How to test this PR?

1. Be on Linux
2. Build an image
3. See that it is chown (ex. ls -l /outputdir)

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -125,6 +125,17 @@ async function fillArchitectures(historyInfo: BootcBuildInfo[]) {
   }
 }
 
+// This will fill the chown function by getting the user and group ID from the OS
+// and filling in the information in the chown input field.
+async function fillChownOption() {
+  try {
+    const gidUid = await bootcClient.getUidGid();
+    buildChown = gidUid;
+  } catch (error) {
+    console.error('Error getting UID and GID:', error);
+  }
+}
+
 async function validate() {
   let prereqs = await bootcClient.checkPrereqs();
   if (prereqs) {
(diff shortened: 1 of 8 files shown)


cdrage:
FYI based on https://github.com/containers/podman-desktop-extension-bootc/pull/684 since it requires certain Linux features to be in for this to work.

cdrage:
Update: Fixing a test, but otherwise ready for review 👍

packages/frontend/src/Build.svelte
```diff
@@ -193,6 +194,7 @@ async function buildBootcImage() {
     type: buildType,
     arch: buildArch,
     filesystem: buildFilesystem,
+    chown: buildChown,
```
deboer-tim, line 197: What do you think about having a default value to change ownership to the current user? I know that's not the 'bootc default', but I think in our context people are more likely to hit the issue like Axel did.

cdrage: I agree and it would be a good idea to implement. Especially being on Linux, users will encounter this issue a lot.

I'll add this in a new commit!

cdrage:
@deboer-tim added chown detection / fill in with a new commit

deboer-tim (approved):

Issue #536, opened 2024-06-11
Initial light mode support


deboer-tim:
### Is your enhancement related to a problem? Please describe

Most light mode support is coming just by reusing components from Podman Desktop, but we're now in the odd place where if we release with the new library components and Podman Desktop starts exposing experimental light mode support, some of the main text and UI in this extension will switch to light mode and expose some issues.

### Describe the solution you'd like

Adopt light mode for most of the UI that we won't get for free from the components - picking up existing style colors from variables.

cdrage:
Split into two issues:

- [ ] https://github.com/containers/podman-desktop-extension-bootc/issues/539
- [ ] https://github.com/containers/podman-desktop-extension-bootc/issues/540

deboer-tim:
Fixing a minor issue with #709, but the rest of this is complete. Marking done and closing.

Issue #376, opened 2024-04-24
Adopt ListItemButtonIcon UI component


jeffmaury:
Depends on https://github.com/containers/podman-desktop/issues/6898

See also https://github.com/containers/podman-desktop/issues/5757

deboer-tim:
Still waiting on https://github.com/containers/podman-desktop/issues/6898.

deboer-tim:
The dependent issue in Podman Desktop to add ListItemButtonIcon to the UI library (https://github.com/podman-desktop/podman-desktop/issues/6898) has been closed as not planned, and the epic closed as well. I would still plan to keep the copy in bootc as a near-direct copy in case that changes, but no point keeping this issue open.

PR #709, opened 2024-08-06
chore: light mode for delete button, remove drop down

### LINKED ISSUE #371: Adopt DropDownMenuItem UI component
Depends on https://github.com/containers/podman-desktop/issues/6918

See also https://github.com/containers/podman-desktop/issues/5757

deboer-tim:
### What does this PR do?

We cannot complete #376 because the issue to provide the component has stalled: https://github.com/containers/podman-desktop/issues/6898

In the meantime we can do two things:
- Pick up light mode support by copying the current code from Podman Desktop.
- Remove the menu property, so that we can remove the unused dependency on the the DropdownMenuItem component. (if we need dropdown support in the future, we should use/require the component)

### Screenshot / video of UI

Minor - in light mode the delete button looks almost disabled before, with color variables it is black.

Related to #376.
Part of #536.
Fixes #371.

### How to test this PR?

Just look at homepage in light mode.


cdrage:
Tested and LGTM!

cdrage (approved):

PR #761, opened 2024-08-28
fix: handle libkrun machines

### LINKED ISSUE #722: Fails root validation check when using macOS + libkrun
### Bug description

Cannot build because it will show the "needs to be root" error.

This is because when using libkrun on macOS, `podman machine ls` will be blank.

We need to:

If we fail validation check and we are on macOS we should:
1. Use `podman system connection ls`
2. Find the default
3. Check the name has `-root`
4. If name does not have root, also check URI that the socket is `/run/podman`
5. Say that it's root podman machine if it passes either.

Luckily it is a validation check and if it it give a false positive, it will fail anyways in `bootc-image-builder` when building, as they ALSO have a validation check to make sure it's root.

### Operating system

mac m1

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

cc @cbr7

jeffmaury:
Fixes #722

### What does this PR do?

Properly handle libkrun machines

### Screenshot / video of UI

N/A

#722

### How to test this PR?

1. Create a libkrun machine
2. Initiate a bootc build

### DIFF
--- a/packages/backend/src/machine-utils.ts
+++ b/packages/backend/src/machine-utils.ts
@@ -21,15 +21,50 @@ import * as os from 'node:os';
 import * as path from 'node:path';
 import * as fs from 'node:fs';
 import { satisfies, coerce } from 'semver';
+import type { ContainerProviderConnection } from '@podman-desktop/api';
+
+function getMachineProviderEnv(connection: extensionApi.ContainerProviderConnection): string {
+  switch (connection.vmType) {
+    case 'wsl':
+    case 'Wsl':
+      return 'wsl';
+    case 'hyperv':
+      return 'hyperv';
+    case 'libkrun':
+    case 'GPU enabled (LibKrun)':
+      return 'libkrun';
+    case 'applehv':
+    case 'default (Apple HyperVisor)':
+      return 'applehv';
+    default:
+      throw new Error('Podman machine not supported on Linux');
+  }
+}
+
+function getPodmanMachineName(connection: ContainerProviderConnection): string {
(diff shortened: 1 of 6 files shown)


cdrage (approved):
https://github.com/user-attachments/assets/cc22e956-7eda-46a1-9d41-d2fb0d4b7254

LGTM!

Thank you for the tests as well, it's unfortunate we do not have a direct Podman Desktop API to handle this, but this solution works.

packages/backend/src/machine-utils.ts
```diff
@@ -21,15 +21,50 @@ import * as os from 'node:os';
 import * as path from 'node:path';
 import * as fs from 'node:fs';
 import { satisfies, coerce } from 'semver';
+import type { ContainerProviderConnection } from '@podman-desktop/api';
+
+function getMachineProviderEnv(connection: extensionApi.ContainerProviderConnection): string {
+  switch (connection.vmType) {
```
cdrage, line 27: maybe we should lowercase vmType before checking so we don't have to add `Wsl` as well?

but just a nitpick, otherwise LGTM.

jeffmaury: I let it on purpose because the 2 values (wsl and Wsl) are there because of a bug in Podman Desktop core that replace the lowercase vmType by the uppercase one in order to get a better displayed value but that's incorrect and in a future only wsl (as well as libkrun) will be sent in 1.13

cdrage: no problem! lgtm

PR #771, opened 2024-09-02
chore(test): move e2e test to pnpm

cbr7:
### What does this PR do?
Updates e2e test to use pnpm and moves to playwright test runner


benoitf (commented):
it looks like it's not only moving the e2e tests to pnpm but the whole project


I'm wondering if we can't split in 2 PR

1. move to playwright (and a minimal set of pnpm (just for e2e)
2. move to pnpm the whole project

cbr7:
@benoitf initially I was only doing the e2e tests, but then I realized that changing the packagemanager setting in package.json in order to get the e2e tests running broke the unit and linter tests

benoitf:
>  the packagemanager setting in package.json in order to get the e2e tests running broke the unit and linter tests

we don't need that AFAIK

cbr7:
@benoitf I was getting this error before, when it wasn't present in the package.json

image

I used the PR from sso extension by Denis as a model and I saw he added the packagemanager setting in the package.json to get it to work.

benoitf:
@cbr7 you can specify a version of pnpm in the github action parameter

also if you specify package.json of podman-desktop I guess it'll find the version of pnpm (as it's only podman desktop that would need pnpm)

cbr7:
Closing for now to not spam with jobs.

Issue #702, opened 2024-08-01
When .iso is selected, show warning cannot build with others selected.


cdrage:
### Bug description

Unfortunately we cannot do multiple types output when iso is selected with raw, qcow2, etc:
```
Build Image Log ----------
ID:     httpd
Image:  quay.io/bootc-extension/httpd
Type:   raw,iso
Folder: /home/axel7083/Downloads
----------
{
  "name": "httpd-bootc-image-builder",
  "Image": "quay.io/centos-bootc/bootc-image-builder:latest-1720185748",
  "Tty": true,
  "HostConfig": {
    "Privileged": true,
    "SecurityOpt": [
      "label=type:unconfined_t"
    ],
    "Binds": [
      "/home/axel7083/Downloads:/output/",
      "/var/lib/containers/storage:/var/lib/containers/storage"
    ]
  },
  "Labels": {
    "bootc.image.builder": "true"
  },
  "Cmd": [
    "quay.io/bootc-extension/httpd:latest",
    "--output",
    "/output/",
    "--local",
    "--type",
    "raw",
    "--type",
    "iso",
    "--target-arch",
    "amd64"
  ]
}
----------
podman \
run \
--rm \
--name \
httpd-bootc-image-builder \
--tty \
--privileged \
--security-opt \
label=type:unconfined_t \
-v \
/home/axel7083/Downloads:/output/ \
-v \
/var/lib/containers/storage:/var/lib/containers/storage \
--label \
bootc.image.builder=true \
quay.io/centos-bootc/bootc-image-builder:latest-1720185748 \
quay.io/bootc-extension/httpd:latest \
--output \
/output/ \
--local \
--type \
raw \
--type \
iso \
--target-arch \
amd64
----------
Trying to pull quay.io/centos-bootc/bootc-image-builder:latest-1720185748...
Getting image source signatures
Copying blob sha256:a9d1ca2b0aab6a93e0a56f5e7955c43138a8e2f46413db6db9dfd8cf3c539412
Copying blob sha256:f0c78caae05f45ada0f664941d8526bcefb371b577e7e8d6265e25a683912807
Copying blob sha256:967f891a5e4fec8b354c88dc83fda687a053d2fc7bfba630eeb45def4348d79f
Copying config sha256:4fbe3c714b191596d6edf8ecc37df0dad149eeb4be17d930f41b35007b26e051
Writing manifest to image destination
Generating manifest manifest-raw-iso.json
[...]

PR #804, opened 2024-09-09
chore: deselect all other options when selecting iso

cdrage:
chore: deselect all other options when selecting iso

### What does this PR do?

* Disables all other options when selecting ISO
* Shows a warning
* Requires
  https://github.com/containers/podman-desktop-extension-bootc/pull/803
  in first

### Screenshot / video of UI

https://github.com/user-attachments/assets/bba03d9c-5f38-40f8-9ea7-0a490a175a0b

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/702

### How to test this PR?

1. Select iso option when building
2. See that it disables everything else


deboer-tim:
To me this is a little odd - we have overall form validation but this appears in a different spot, and selecting iso (even accidentally) clears the other selections. Couldn't we just do this as a regular validation? When iso + anything else is selected, build is disabled and validation says something like 'You cannot build ISO at the same time as any other type'.

cdrage:
> To me this is a little odd - we have overall form validation but this appears in a different spot, and selecting iso (even accidentally) clears the other selections. Couldn't we just do this as a regular validation? When iso + anything else is selected, build is disabled and validation says something like 'You cannot build ISO at the same time as any other type'.

Sure, we could do that instead, can just validate and say that "unable to pick iso" like you mentioned before if others are selected.

Issue #813, opened 2024-09-10
Add initial "launch VM" implementation with terminal


cdrage:
### Is your enhancement related to a problem? Please describe

See: https://github.com/containers/podman-desktop-extension-bootc/issues/752 

### Describe the solution you'd like

Part of the epic to add the initial launch VM implementation with terminal access.

PR #865, opened 2024-09-27
feat: add launch VM button

cdrage:
feat: add launch VM button

### What does this PR do?

* Using QEMU we add a feature to "Launch VM" in the background
* Uses websockets, qemu as well as our xterm.js library to achieve this
* Launches in "snapshot" mode so no data is written to .raw file so the
  file can be easily re-used

### Screenshot / video of UI

https://github.com/user-attachments/assets/ab3f0fbf-fd80-40b7-b499-30d1f10b66c1

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/813

### How to test this PR?

1. Be on macOS silicon
2. `brew install qemu`
3. Build a bootc container image
4. Press launch VM button in actions bar


cdrage:
This is an **initial** implementation and meant for experimenting!

Feel free to test it out @deboer-tim @slemeur 

We can further discuss this during the week!

deboer-tim:
Comments from my initial testing:
- That's a lot of text on the page. I wonder if we can detect if qemu is installed and give shorter messages, shorten text, or say there are caveats and move those to docs?
- Do we hide the page for non-raw disk images? Feels like it might be annoying to always have it there when you can't use it, but is also good advertising.
- I get errors like `Error stopping VM:  sh: line 0: kill: (97942) - No such process` in the log, even before I've tried to start a VM.
- Woo hoo! It worked perfectly, and was so nice to see this function (again).
- The `Connection opened` label is maybe correct, but I think something like ' VM running' might be more intuitive.
- I wanted to check something from the image I was using... unfortunate that you can't pop away for a minute.
- I know leaving the page will stop it, but I felt like I should have a Stop button too. Or conversely, if I accidentally click out before stopping it might be nice to have a warning?

cdrage:
> Comments from my initial testing:
> 
> * That's a lot of text on the page. I wonder if we can detect if qemu is installed and give shorter messages, shorten text, or say there are caveats and move those to docs?

Agreed, let's launch it and maybe just link to docs near the status icon to give more information?

> * Do we hide the page for non-raw disk images? Feels like it might be annoying to always have it there when you can't use it, but is also good advertising.

I don't think we should hide, but maybe an empty page saying "must build raw image, no raw image found"?

> * I get errors like `Error stopping VM:  sh: line 0: kill: (97942) - No such process` in the log, even before I've tried to start a VM.

Intentional! Sorry!

> * Woo hoo! It worked perfectly, and was so nice to see this function (again).
> * The `Connection opened` label is maybe correct, but I think something like ' VM running' might be more intuitive.

Will fix.

> * I wanted to check something from the image I was using... unfortunate that you can't pop away for a minute.

I agree :( I think we should have that in a follow up issue / PR though.

> * I know leaving the page will stop it, but I felt like I should have a Stop button too. Or conversely, if I accidentally click out before stopping it might be nice to have a warning?

cdrage:
https://github.com/user-attachments/assets/841b20ef-ba29-481a-8d6c-797b9a431d9f

@deboer-tim Ready for another review! I have done the following:

- No more button to launch VM, it will insta-launch
- Added prereq checks for mac + QEMU
- Removed the error logging (except for when a VM is shut down, still working on that).
- Fixed connection closed / VM status labels


Feel free to test it out and let me know if this works and I can begin adding some unit tests so we can get this initial POC in.

cdrage:
Ready for review with the exception of unit tests being added later.

This is marked as experimental (as per the tab) and only shown for macOS users.

packages/shared/src/BootcAPI.ts
```diff
@@ -36,9 +37,13 @@ export abstract class BootcApi {
   abstract generateUniqueBuildID(name: string): Promise<string>;
   abstract openLink(link: string): Promise<void>;
   abstract isLinux(): Promise<boolean>;
+  abstract isMac(): Promise<boolean>;
   abstract getUidGid(): Promise<string>;
   abstract loadLogsFromFolder(folder: string): Promise<string>;
   abstract getConfigurationValue(config: string, section: string): Promise<unknown>;
+  abstract launchVM(folder: string, architecture: string): Promise<void>;
+  abstract readFromClipboard(): Promise<string>;
+  abstract stopVM(): Promise<void>;
```
axel7083, line 46: only one VM can run ?

cdrage: at the moment yes until #876 is implemented, didn't want to make this POC too big.

we also have to make sure if it's still running in the background, etc.

packages/backend/src/launch-vm.ts
```diff
@@ -0,0 +1,185 @@
+import path from 'node:path';
```
axel7083, line 1: this file is very specific to Mac with a lot of constants specific to this system, how would this scale with https://github.com/containers/podman-desktop-extension-bootc/issues/878 and https://github.com/containers/podman-desktop-extension-bootc/issues/879

packages/backend/src/machine-utils.ts
```diff
@@ -174,6 +174,11 @@ export function isLinux(): boolean {
   return linux;
 }
 
+const darwin = os.platform() === 'darwin';
+export function isMac(): boolean {
+  return darwin;
+}
+
```
axel7083, line 181: You could use

```ts
import { env } from '@podman-desktop/api'
env.isMac
```

packages/backend/src/launch-vm.ts
cdrage: One step at a time to be honest! When implementing Linux support, it'll be more modular / the file will be way less specific.

packages/backend/src/machine-utils.ts
cdrage: opened up: https://github.com/containers/podman-desktop-extension-bootc/issues/881

cdrage: updated the code as well :)

axel7083 (approved):
I do not have a Mac so my review is Code-wise only ! Great work

jeffmaury (commented):
Out of curiosity, is QEMU a strong requirement ? Why not using vfkit to launch the machine ?

cdrage:
> Out of curiosity, is QEMU a strong requirement ? Why not using vfkit to launch the machine ?

Vfkit does not have websockets or tty support (they are working on it, but it's been on hold and not merged in yet for 6+ months) to allow for easy connection to it through xterm. 

Luckily once that support is in, we can easily swap in the commands requirements into bootc.

packages/backend/src/launch-vm.ts
```diff
@@ -0,0 +1,185 @@
...
+// eslint-disable-next-line sonarjs/publicly-writable-directories
+const pidFile = '/tmp/qemu-podman-desktop.pid';
+
+// Must use "homebrew" qemu binaries on macOS
+// as they are found to be the most stable and reliable for the project
+// as well as containing the necessary "edk2-aarch64-code.fd" file
+// it is not advised to use the qemu binaries from qemu.org due to edk2-aarch64-code.fd not being included.
+const macQemuArm64Binary = '/opt/homebrew/bin/qemu-system-aarch64';
+const macQemuArm64Edk2 = '/opt/homebrew/share/qemu/edk2-aarch64-code.fd';
+const macQemuX86Binary = '/opt/homebrew/bin/qemu-system-x86_64';
```
benoitf, line 17: does it mean, it'll work only on mac/Silicon and not mac/Intel ?

packages/backend/src/launch-vm.ts
```diff
@@ -0,0 +1,185 @@
+import path from 'node:path';
+import * as extensionApi from '@podman-desktop/api';
```
benoitf, line 2: missing copyright headers

packages/backend/src/launch-vm.ts
benoitf: also we recommend to not install podman using brew but here it'll install qemu using brew ?!

packages/backend/src/launch-vm.ts
```diff
@@ -0,0 +1,185 @@
...
+  // Will ONLY work with RAW images located at image/disk.raw which is the default output location
+  const diskImage = path.join(folder, rawImageLocation);
+
+  // Check to see that the disk image exists before continuing
+  if (!fs.existsSync(diskImage)) {
+    throw new Error(`Raw disk image not found: ${diskImage}`);
+  }
+
+  // Before launching, make sure that we stop any previously running VM's and ignore any errors when stopping
+  //await stopVM();
```
benoitf, line 40: ?

packages/backend/src/launch-vm.ts
benoitf: I concur to the @axel7083 comment here, it should be designed from the beginning as pluggable runtime, not a list of functions

Like you plug qemu/mac support, then you could plug vfkit/mac, etc.

Having like tons of functions in the same file will make it a pure nightmare to maintain as it's not scoped.
I don't see why architecture needs to be sent to each function, it's something stateful from the runner

benoitf (dismissed):
need to revisit the design to be more pluggable

also probably needs to have a Class/instance/state per runner as here we're passing tons of useless parameters

packages/backend/src/launch-vm.ts
cdrage: @benoitf Yes, it will only work with mac M1 as that's what we're supporting at the moment.

The reasoning for brew install is that is the recommended install pattern by podman-bootc (https://github.com/containers/podman-bootc), vfkit as well as the qemu project: https://www.qemu.org/download/#macos

Downloading the binaries directly would not be good as there are *a lot* that's needed / added (qemu-img, qemu-io, qemu-user-static, etc.):

```sh
▶ qemu-
qemu-edid                 qemu-system-loongarch64   qemu-system-riscv32
qemu-img                  qemu-system-m68k          qemu-system-riscv64
qemu-io                   qemu-system-microblaze    qemu-system-rx
qemu-nbd                  qemu-system-microblazeel  qemu-system-s390x
qemu-storage-daemon       qemu-system-mips          qemu-system-sh4
qemu-system-aarch64       qemu-system-mips64        qemu-system-sh4eb
qemu-system-alpha         qemu-system-mips64el      qemu-system-sparc
qemu-system-arm           qemu-system-mipsel        qemu-system-sparc64
qemu-system-avr           qemu-system-nios2         qemu-system-tricore
qemu-system-cris          qemu-system-or1k          qemu-system-x86_64
qemu-system-hppa          qemu-system-ppc           qemu-system-xtensa
qemu-system-i386          qemu-system-ppc64         qemu-system-xtensaeb
```

cdrage:
> need to revisit the design to be more pluggable
> 
> also probably needs to have a Class/instance/state per runner as here we're passing tons of useless parameters

ACK, I'll refactor the launch-vm.ts part to make it more pluggable for future implementations (linux, windows, etc.)

packages/shared/src/BootcAPI.ts
```diff
@@ -21,6 +21,7 @@ import type { ImageInfo, ImageInspectInfo, ManifestInspectInfo } from '@podman-d
 
 export abstract class BootcApi {
   abstract checkPrereqs(): Promise<string | undefined>;
+  abstract checkVMLaunchPrereqs(folder: string, architecture: string): Promise<string | undefined>;
```
deboer-tim, line 24: Both backend and frontend know about list of BootcBuildInfos (i.e. disk images), but frontend grabs folder and arch from one of these to check if it can be launched, then to actually launchVM(). Seems like we could use the BootBuildInfo or its id directly instead? i.e. can I launch disk image X / run disk image Y.

packages/backend/src/vm-manager.ts
```diff
@@ -0,0 +1,201 @@
...
+const websocketPort = '45252';
+const rawImageLocation = 'image/disk.raw';
+
+export default class VMManager {
+  private folder: string;
+  private architecture: string;
+
+  // Only values needed is the location of the VM file as well as the architecture of the image that
+  // will be used.
+  constructor(folder?: string, architecture?: string) {
```
deboer-tim, line 45: Ditto to comment in BootcAPI.ts, it's not clear why this doesn't just use the BootcBuildInfo or id.

Also would be simpler (and closer to future API?) to create only a single instance of VMManager and pass the arg to each function, vs currently creating an instance of VMManager for each API call.

deboer-tim (commented):
As before, testing worked without any issues.

I don't love auto-launching a 'large' VM, but that's fine for first pass. The main drawback is the VM getting reset when you leave the tab, but again that is expected for this first step.

deboer-tim:
Thanks for the changes! I was also suggesting instead of:
`abstract launchVM(folder: string, architecture: string): Promise<void>;`
API could be:
`abstract launchVM(diskImageId: string): Promise<void>;`
but please go ahead with wrapping up tests.

cdrage:
> Thanks for the changes! I was also suggesting instead of: `abstract launchVM(folder: string, architecture: string): Promise<void>;` API could be: `abstract launchVM(diskImageId: string): Promise<void>;` but please go ahead with wrapping up tests.

Done! I use BuildImageInfo for that section now because who knows what other information we may be passing in the future now.

cdrage:
Tests added and ready for review 👍 

@benoitf @deboer-tim

packages/frontend/src/lib/disk-image/DiskImageActions.svelte
```diff
@@ -17,7 +20,20 @@ async function deleteBuild(): Promise<void> {
 async function gotoLogs(): Promise<void> {
   router.goto(`/disk-image/${btoa(object.id)}/build`);
 }
+
+async function gotoVM(): Promise<void> {
+  router.goto(`/details/${btoa(object.id)}/vm`);
```
deboer-tim, line 25: I think this needs to be updated to `/disk-image/` after the nav change.

deboer-tim (approved):
Thanks for all the improvements, @cdrage. One minor url change noted but otherwise let's get this in and get some feedback!

cdrage:
@deboer-tim Updated and confirm link works now! 

Agreed, let's get this in even if it's in nightly and get some feedback :)


https://github.com/user-attachments/assets/32920229-a2f2-4892-bf62-4184d4c0fd36

packages/backend/src/api-impl.ts
```diff
@@ -60,6 +65,31 @@ export class BootcApiImpl implements BootcApi {
...
+      let errorMessage: string;
+      if (e instanceof Error) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        errorMessage = `${e.message} ${'stderr' in e ? (e as any).stderr : ''}`;
+      } else {
+        errorMessage = String(e);
+      }
+      await this.notify(Messages.MSG_VM_LAUNCH_ERROR, { success: '', error: errorMessage });
+    }
+    return Promise.resolve();
```
benoitf, line 85: do not need to return anything
it's an async method being void so just remove

packages/backend/src/api-impl.ts
```diff
@@ -60,6 +65,31 @@ export class BootcApiImpl implements BootcApi {
...
+  }
+
+  // Stop VM by pid file on the system
+  async stopCurrentVM(): Promise<void> {
+    return await new VMManager().stopCurrentVM();
```
benoitf, line 90: do not return await here, return the promise

packages/backend/src/machine-utils.ts
```diff
@@ -174,6 +175,18 @@ export function isLinux(): boolean {
...
+  return env.isMac;
+}
+
+export function isArm(): boolean {
+  return os.arch() === 'arm64';
+}
+
+export function isX86(): boolean {
+  return os.arch() === 'x64';
+}
```
benoitf, line 188: maybe file an issue to add arch on the env

packages/backend/src/vm-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+const websocketPort = '45252';
+const rawImageLocation = 'image/disk.raw';
+
+export default class VMManager {
+  private build: BootcBuildInfo;
+
+  // Only values needed is the location of the VM file as well as the architecture of the image that
+  // will be used.
+  constructor(build?: BootcBuildInfo) {
+    this.build = build!;
```
benoitf, line 46: please don't do that

it can't be optional and then you're saying there is a value with !

it will make unexpected runtime failures

packages/backend/src/vm-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    if (!this.isArchitectureSupported()) {
+      return `Unsupported architecture: ${this.build.arch}`;
+    }
+
+    // Future support for Mac Intel, Linux ARM, Linux X86 and Windows ARM, Windows X86 to be added here.
+    if (isMac() && isArm()) {
+      return this.checkMacPrereqs();
+    } else {
```
benoitf, line 99: I think it doesn't scale well there, the idea was to have multiple instances

like you have an abstract base, then a MacArm, a MacIntel that has a preReqs method to fill
it's not the client that needs to do the if if at each call

packages/backend/src/vm-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+      return 'Unsupported OS. Only MacOS Silicon is supported.';
+    }
+  }
+
+  private getDiskImagePath(): string {
+    return path.join(this.build.folder, rawImageLocation);
+  }
+
+  private isArchitectureSupported(): boolean {
+    return this.build.arch === 'amd64' || this.build.arch === 'arm64';
```
benoitf, line 109: should be part of the Mac implementation

packages/backend/src/vm-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+      return `QEMU arm64 edk2-aarch64-code.fd file not found at ${macQemuArm64Edk2}. ${installDisclaimer}`;
+    }
+    return undefined;
+  }
+
+  // Supported: MacOS Silicon
+  // Unsupported: MacOS Intel, Linux, Windows
+  private generateLaunchCommand(diskImage: string): string[] {
+    // Future support for Mac Intel, Linux ARM, Linux X86 and Windows ARM, Windows X86 to be added here.
+    if (isMac() && isArm()) {
```
benoitf, line 130: if we had proper method in macArm implementation, we would not need the is and is

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
```diff
@@ -0,0 +1,309 @@
...
+  bootcClient.launchVM(build);
+
+  // Initialize the terminal so it awaits the websocket connection.
+  await initTerminal();
+
+  // To avoid a blank terminal wait until terminal has logs and and then show it
+  // logs.terminal.buffer.normal will contain the "ascii cursor" with a value of 1 until there is more logs.
+  // we wait until buffer.normal.length is more than 1.
+  while (logsTerminal.buffer.normal.length < 1) {
+    await new Promise(resolve => setTimeout(resolve, 500));
```
benoitf, line 191: it should have a safe exit (max-retries) to avoid infinite loop

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
```diff
@@ -0,0 +1,309 @@
...
+// be present in the terminal.
+async function resetTerminalTheme(): Promise<void> {
+  // Wait until we have more than 100 lines in the buffer before resetting the terminal
+  while (logsTerminal.buffer.normal.length < 100) {
+    await new Promise(resolve => setTimeout(resolve, 500));
+  }
+
+  // Reset the terminal
+  logsTerminal.reset();
+}
```
benoitf, line 216: it doesn't reset the theme, it reset the terminal

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
```diff
@@ -0,0 +1,309 @@
...
+        if (socket !== undefined) {
+          const encoder = new TextEncoder();
+          const binaryData = encoder.encode(clipboardText);
+          socket.send(binaryData.buffer);
+        }
+      });
+    }
+    return true;
+  });
+}
```
benoitf, line 174: why do we need that, I can copy/paste in Terminal of Podman Desktop without this

benoitf (commented):
> @[cdrage](https://github.com/cdrage) cdrage dismissed [benoitf]

please do not dismiss the review of someone without asking him first

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
cdrage: The reason is because we are using the special websockets mode for addon-attach https://www.npmjs.com/package/xterm-addon-attach and unfortunately we have to "disable stdin" since it's more meant for viewing websocket connections rather than interaction.

That's why we have to add a custom event handler for copying and pasting code as well as interpret each keystroke (see key section code).

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
cdrage: That's intentional as it will clear any ASCII background characters in the log output. Tried many other alternative ways (including "setting the theme again") and nothing works.

It's a bug with websockets mode and using the host hardware for mac support and luckily a very minor edge case.

packages/backend/src/machine-utils.ts
cdrage: https://github.com/containers/podman-desktop/issues/9560

Issue #808, opened 2024-09-09
KNOWN ISSUE: Unable to cross-arch build 


cdrage:
### Bug description

If you have output issues such as:

```
error: Performing deployment: Creating importer: Function not implemented (os error 38)
Traceback (most recent call last):
  File "/run/osbuild/bin/org.osbuild.ostree.deploy.container", line 72, in <module>
    r = main(stage_args["tree"],
  File "/run/osbuild/bin/org.osbuild.ostree.deploy.container", line 67, in main
    ostree_container_deploy(tree, inputs, osname, target_imgref, kopts)
  File "/run/osbuild/bin/org.osbuild.ostree.deploy.container", line 41, in ostree_container_deploy
    ostree.cli("container", "image", "deploy",
  File "/run/osbuild/lib/osbuild/util/ostree.py", line 205, in cli
    return subprocess.run(["ostree"] + args,
  File "/usr/lib64/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,

manifest - failed 
```

There are current issues upstream regarding building the default stream9 image or cross-arch builds that you may encounter:
- https://github.com/osbuild/bootc-image-builder/issues/619
- https://github.com/osbuild/bootc-image-builder/issues/641

### Operating system

all OS'

### Version

next (development version)

cdrage:
UPDATE there's a solution now!

With Rosetta disabled and the instructions from https://github.com/osbuild/bootc-image-builder/issues/619#issuecomment-2360569582 I am now able to build.

cdrage:
Marked this as Done as part of sprint planning as we now have the document on the README: https://github.com/containers/podman-desktop-extension-bootc?tab=readme-ov-file#known-issues

But I will leave this open in case someone else stumbles upon this issue.

cdrage:
Closing this as it's been left open since Oct 2024 and we have multiple other issues with regards to cross-arch building (and a section on our README)

PR #889, opened 2024-10-07
docs: add known issue cross-arch building

cdrage:
docs: add known issue cross-arch building

### What does this PR do?

* Adds documentation regarding cross-arch building
* Docs are meant as an intermediary until podman machine is patched with
  the new QEMU calls
* When QEMU is patched, the only thing needed is to "disable rosetta"

### Screenshot / video of UI

N/A

References https://github.com/containers/podman-desktop-extension-bootc/issues/808

### How to test this PR?

Follow documentation instructions on how to cross-arch build again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Easily go from container to VM / ISO-on-a-USB / RAW image!
 - [Requirements](#requirements)
 - [Installation](#installation)
 - [Usage](#usage)
+- [Preferences](#preferences)
+- [Known Issues](#known-issues)
 - [Contributing](#contributing)
 
 ## Technology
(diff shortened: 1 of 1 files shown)


gastoner (approved):

deboer-tim (commented):
First 5 lines, linking to upstream bug - excellent.

Next bit - I'm not sure we should be explaining the cause of the bug here or how it will be resolved, would prefer to see that upstream.

'follow these steps to resolve' followed by 'preliminary steps' - It's not clear to me as a Mac user whether this will fix the problem or these are steps I will need to take before the issue is resolved. If the latter, once the fix comes will I still need to take these steps? (and again, feels like this should be upstream)

As a Windows user, I'm assuming this issue applies to me? But 'follow these steps to resolve' only list Mac, would be nice to confirm platforms and that we don't have steps for others at the moment.

cdrage:
> First 5 lines, linking to upstream bug - excellent.
> 
> Next bit - I'm not sure we should be explaining the cause of the bug here or how it will be resolved, would prefer to see that upstream.

Upstream fix has already been applied for QEMU.

However, Rosetta must be disabled even after the QEMU fix goes in, so we'll still have to have this section :(. We are a special edge case that uses these openat2 system calls which are causing issues in the macOS kernel with Rosetta. 

> 
> 'follow these steps to resolve' followed by 'preliminary steps' - It's not clear to me as a Mac user whether this will fix the problem or these are steps I will need to take before the issue is resolved. If the latter, once the fix comes will I still need to take these steps? (and again, feels like this should be upstream)

Updated the headers to indicate more clearly it's for macOS and Windows.

> 
> As a Windows user, I'm assuming this issue applies to me? But 'follow these steps to resolve' only list Mac, would be nice to confirm platforms and that we don't have steps for others at the moment.

Updated the steps! :)

jeffmaury (commented):
I'm wondering if we can't use the RPM repo URL instead of direct link to the RPM that will ensure if a new build is created we will get it

cdrage:
> I'm wondering if we can't use the RPM repo URL instead of direct link to the RPM that will ensure if a new build is created we will get it

No new builds will be created since it's already been merged into QEMU upstream, so just waiting for podman to update their machine images / qemu to do a new release as well.

deboer-tim (approved):

PR #913, opened 2024-10-11
feat: go directly to build logs on successful build

### LINKED ISSUE #872: Go directly to build logs when starting a build

When you start a disk image build we currently open a screen to tell you the build is started, and then you click a button to return to the homepage. If you want to see the build logs, you need to click on the build.

Now that we have a page showing build logs, it would make more sense to jump directly to that page. If the user wants to see progress that's where it is; if they don't it's still only one click to get back to the homepage (or elsewhere).

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

When a build is started successfully, go directly to the build logs.

Switches to use the store instead of calling bootcClient.listHistoryInfo() repeatedly, less for performance/reuse and more to ensure that the store has been updated before switching to the build logs. Otherwise we may try to switch before the store has been updated.

Noticed a pattern in a few other tests of clicking and then 'await new Promise', fixed these to await the userEvent while I was here.

### Screenshot / video of UI

https://github.com/user-attachments/assets/5c392ef0-e522-4d8b-978a-4e8e791ab0a2

Fixes #872.

### How to test this PR?

Start a build and make sure it automatically switches to the Build Logs instead of the previous page.

### DIFF
--- a/packages/frontend/src/Build.spec.ts
+++ b/packages/frontend/src/Build.spec.ts
@@ -24,6 +24,9 @@ import Build from './Build.svelte';
 import type { BootcBuildInfo } from '/@shared/src/models/bootc';
 import type { ImageInfo, ImageInspectInfo, ManifestInspectInfo } from '@podman-desktop/api';
 import { bootcClient } from './api/client';
+import { router } from 'tinro';
+import userEvent from '@testing-library/user-event';
+import { historyInfo } from './stores/historyInfo';
 
 const mockHistoryInfo: BootcBuildInfo[] = [
   {
@@ -100,6 +103,8 @@ vi.mock('./api/client', async () => {
       inspectImage: vi.fn(),
       inspectManifest: vi.fn(),
       isLinux: vi.fn().mockImplementation(() => mockIsLinux),
+      generateUniqueBuildID: vi.fn(),
+      buildImage: vi.fn(),
     },
     rpcBrowser: {
       subscribe: () => {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/Build.spec.ts
```diff
@@ -745,28 +763,52 @@ test('select anaconda-iso and qcow2 and expect validation error to be shown', as
...
+  vi.mocked(bootcClient.buildExists).mockResolvedValue(true);
+  vi.mocked(bootcClient.inspectImage).mockResolvedValue(mockImageInspect);
+
+  // mock unique id so that it matches an image in the history ('new' build exists already)
+  vi.mocked(bootcClient.generateUniqueBuildID).mockReturnValue(Promise.resolve(mockHistoryInfo[0].id));
+
+  render(Build);
+
+  // Wait until children length is 2 meaning it's fully rendered / propagated the changes
+  while (screen.getByLabelText('image-select')?.children.length !== 2) {
```
jeffmaury, line 796: Why don't you use vi.waitUntil

axel7083: vi.waitFor

deboer-tim: Yes, that is better. Was just copying what was there but switched all 9 instances to vi.waitFor(). Seems like there should be a more succinct version but it escapes me atm.

cdrage (approved):
LGTM. Tested and the code looks good too.

It's good that we wait until it's available in the history and then change pages.

PR #915, opened 2024-10-11
chore: always pass empty body to notify message

### LINKED ISSUE #914: Fix incompatible message error in logs
### Bug description

In the logs I sometimes see errors like this:
`Received incompatible message. {id: 'history-update', body: ''}`

### Operating system

macOS 15.0

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Confirmed the error isn't impacting function at all, but use an empty object instead of an empty string to avoid errors in the log.

The code in api-impl.ts had an optional parameter that was always passed, just make the parameter mandatory to avoid the same problem in the future.

### Screenshot / video of UI

N/A

Fixes #914.

### How to test this PR?

PR checks, or try creating/deleting builds and make sure the notifier error is no longer in the logs.

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -275,7 +275,7 @@ export class BootcApiImpl implements BootcApi {
   // The API does not allow callbacks through the RPC, so instead
   // we send "notify" messages to the frontend to trigger a refresh
   // this method is internal and meant to be used by the API implementation
-  protected async notify(msg: string, body: unknown): Promise<void> {
+  protected async notify(msg: string, body: unknown = {}): Promise<void> {
     await this.webview.postMessage({
       id: msg,
       // Must pass in an empty body to satisfy the type system, if it is undefined, this fails.
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -275,11 +275,11 @@ export class BootcApiImpl implements BootcApi {
   // The API does not allow callbacks through the RPC, so instead
   // we send "notify" messages to the frontend to trigger a refresh
   // this method is internal and meant to be used by the API implementation
-  protected async notify(msg: string, body: unknown = {}): Promise<void> {
+  protected async notify(id: string, body: unknown = {}): Promise<void> {
+    // Must pass in an empty body, if it is undefined this fails
     await this.webview.postMessage({
-      id: msg,
-      // Must pass in an empty body to satisfy the type system, if it is undefined, this fails.
-      body: body,
+      id,
+      body,
     });
   }
 }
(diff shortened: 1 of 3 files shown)


gastoner:
I'm not able to create a build, I always get 125 error (but probably just wrong options), otherwise I don't see the notifier

packages/backend/src/api-impl.ts
```diff
@@ -275,11 +275,11 @@ export class BootcApiImpl implements BootcApi {
   // The API does not allow callbacks through the RPC, so instead
   // we send "notify" messages to the frontend to trigger a refresh
   // this method is internal and meant to be used by the API implementation
-  protected async notify(msg: string, body?: unknown): Promise<void> {
+  protected async notify(msg: string, body: unknown): Promise<void> {
```
benoitf, line 278: maybe you should use the following pattern
```suggestion
  protected async notify(msg: string, body: unknown = {}): Promise<void> {
```

deboer-tim: Excellent point, fixed in new commit.

benoitf (approved):

packages/backend/src/api-impl.ts
```diff
@@ -275,11 +275,11 @@ export class BootcApiImpl implements BootcApi {
   // The API does not allow callbacks through the RPC, so instead
   // we send "notify" messages to the frontend to trigger a refresh
   // this method is internal and meant to be used by the API implementation
-  protected async notify(msg: string, body?: unknown): Promise<void> {
+  protected async notify(msg: string, body: unknown = {}): Promise<void> {
     await this.webview.postMessage({
       id: msg,
       // Must pass in an empty body to satisfy the type system, if it is undefined, this fails.
-      body: body ?? '',
+      body: body,
```
benoitf, line 282: can be just `body` rather than `body: body`

packages/backend/src/history/historyNotifier.ts
```diff
@@ -39,7 +39,7 @@ export class HistoryNotifier implements Disposable {
     await this.webview.postMessage({
       id: Messages.MSG_HISTORY_UPDATE,
       // Must pass in an empty body to satisfy the type system, if it is undefined, this fails.
-      body: '',
+      body: {},
```
benoitf, line 42: can be removed (no body parameter)

packages/backend/src/api-impl.ts
deboer-tim: Good point, simplified this and renamed `msg` to `id` so that it can do the same.

packages/backend/src/history/historyNotifier.ts
benoitf: yes, exact, reading I thought we had access to the notify method (and calling that notify method)

deboer-tim: There are unfortunately two places bootc backend calls webview.postMessage(), both need to pass a body and I was just trying to fix both calls here. IMHO this instance should call api-imp.ts notify instead of webview directly, but wasn't trying to fix that tech debt at the same time.

benoitf: sure, it was me that read incorrectly

cdrage (approved):

Issue #895, opened 2024-10-09
Implementation: How to get examples from outside sources (bootc fedora examples)


cdrage:
### Is your feature request related to a problem? Please describe

Depends on https://github.com/containers/podman-desktop-extension-bootc/issues/894

### Describe the solution you'd like

Implementation on how to retrieve examples

PR #920, opened 2024-10-15
feat: backend for examples catalog

cdrage:
feat: backend for examples catalog

### What does this PR do?

Adds the basic backend functionality for the examples catalog.

This adds a JSON file that lists all the currently available examples.

### Screenshot / video of UI

NOT INCLUDED IN PR. But part of this UI implementation:

https://github.com/user-attachments/assets/9205c5f3-40a6-4def-a4ee-2653c4165c25

Part of https://github.com/containers/podman-desktop-extension-bootc/issues/895

### How to test this PR?

N/A

### DIFF
--- a/packages/backend/src/api-impl.spec.ts
+++ b/packages/backend/src/api-impl.spec.ts
@@ -48,5 +48,9 @@ test('getExamples should return examplesCatalog', async () => {
 
   // When running get Examples we should return the examplesCatalog (it's exported)
   const result = await bootcApi.getExamples();
+  // Check that the examples and categories are NOT empty
+  expect(result.examples.length).not.toBe(0);
+  expect(result.categories.length).not.toBe(0);
+
   expect(result).toEqual(examplesCatalog as ExamplesList);
 });
--- a/packages/backend/src/api-impl.spec.ts
+++ b/packages/backend/src/api-impl.spec.ts
@@ -1,0 +1,52 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { expect, test, vi } from 'vitest';
+import examplesCatalog from '../assets/examples.json';
+import type { ExamplesList } from '/@shared/src/models/examples';
+import { BootcApiImpl } from './api-impl';
+import type * as podmanDesktopApi from '@podman-desktop/api';
+
+vi.mock('@podman-desktop/api', async () => {
+  return {
+    window: {
+      showErrorMessage: vi.fn(),
+    },
+    containerEngine: {
+      listImages: vi.fn(),
+    },
+    env: {
+      openExternal: vi.fn(),
+      createTelemetryLogger: vi.fn(),
+    },
+  };
+});
+
+test('getExamples should return examplesCatalog', async () => {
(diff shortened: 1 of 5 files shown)


packages/shared/src/models/examples.ts
```diff
@@ -0,0 +1,59 @@
...
+export type ArchitectureType = 'arm64' | 'amd64';
+
+export interface Example {
+  id: string;
+  name: string;
+  categories: string[];
+  description: string;
+  repository: string;
+
+  // List of available architectures to pull wtih the image.
```
deboer-tim, line 35: Typo 'with'.

packages/shared/src/models/examples.ts
```diff
@@ -0,0 +1,59 @@
...
+  tag?: string;
+
+  // Optional reference version for the example
+  ref?: string;
+
+  // Will assume 'Containerfile' in the root of the repo.
+  basedir?: string;
+
+  // Marks if it is already pulled or not
+  pulled?: boolean;
```
deboer-tim, line 51: These are the only cases of 'state' where everything else comes from the file. Feels a bit odd here since I can't review how the data is merged, wondering if it should be left off until a PR that would set it? Maybe better to have a separate object for state?

Second, pulled/pulling can't both be true so maybe should be an enum? But then maybe we want a 'building disk image' state? Again, might be easier to remove for now and review in context.

packages/shared/src/models/examples.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// validator for certain types arm / amd64
+export type ArchitectureType = 'arm64' | 'amd64';
+
+export interface Example {
+  id: string;
+  name: string;
+  categories: string[];
+  description: string;
+  repository: string;
```
deboer-tim, line 33: The first few are obvious, but maybe better to have a comment on repo to be overly clear if it is image or git repo.

packages/shared/src/models/examples.ts
```diff
@@ -0,0 +1,59 @@
...
+  // List of available architectures to pull wtih the image.
+  architectures?: ArchitectureType[];
+
+  // container image to pull, some may not be available... so make this optional.
+  image?: string;
+
+  // tag, else we will assume 'latest'
+  tag?: string;
+
+  // Optional reference version for the example
```
deboer-tim, line 44: Nice this one says Optional, but it's the only one that doesn't say what the default behaviour is. All of these optional parameters are phrased a bit differently, would be nice to make them all consistent. e.g. like 'X. Defaults to Y', don't use 'will assume'.

packages/backend/src/api-impl.ts
```diff
@@ -42,6 +44,10 @@ export class BootcApiImpl implements BootcApi {
     this.webview = webview;
   }
 
+  async getExamples(): Promise<ExamplesList> {
+    return examplesCatalog as ExamplesList;
```
deboer-tim, line 48: Could use a test. I know it's not testing much, but just to confirm that we get examples loaded here.

Maybe could confirm some basic hygiene like 'all examples have descriptions with length > 20' or repo/image strings look valid?

deboer-tim (changes requested):
Looks good, just requesting some minor improvements.

packages/shared/src/models/examples.ts
cdrage: Honestly it would be good just to remove this as "tag" would be used for the image, and we are not pulling / using references yet for git... Could revisit in the future if we add the functionality to download the Containerfile locally / git checkout.

cdrage:
Added some tests and updated based on the review! @deboer-tim

packages/shared/src/models/examples.ts
cdrage: Done!

packages/backend/src/api-impl.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    storagePath: '/fake-path',
+  } as unknown as podmanDesktopApi.ExtensionContext;
+
+  const webviewMock = {} as podmanDesktopApi.Webview;
+
+  const bootcApi = new BootcApiImpl(extensionContextMock, webviewMock);
+
+  // When running get Examples we should return the examplesCatalog (it's exported)
+  const result = await bootcApi.getExamples();
+  expect(result).toEqual(examplesCatalog as ExamplesList);
```
deboer-tim, line 51: This is a really simple test. Could you at least check examples length and categories length is > 0 too?

deboer-tim (approved):

Issue #886, opened 2024-10-04
Sub-navigation and initial landing page


deboer-tim:
As a precursor to #863, we need sub-navigation in the bootc extension and (so that we have more than one page) we will start with the current welcome.

- Add sub navigation.
- Move current welcome to a dedicated (the default) page
- Add a more traditional empty screen (ideally still with actions) to the disk images page.

PR #934, opened 2024-10-16
feat: navigation and dashboard

deboer-tim:
### What does this PR do?

Introduces left navigation for bootc. The initial Dashboard is just the empty screen that we had before (but now always visible), and Disk Images is the main screen we had after you had created an image, renamed and with a more typical empty screen. This sets bootc up for adding additional pages in the future, most likely Samples and a filtered bootable Images page.

Most page urls had to be updated, and I switched to vi.waitFor() for the tests I updated.

### Screenshot / video of UI

https://github.com/user-attachments/assets/5c5a21db-6642-42cc-bb04-ce8280f7699d

Most of #886, but not marking as fixed since there is some renaming & cleanup that I'd like to do on other files but didn't want to complicate this PR further.

### How to test this PR?

Unit tests updated. Go to every page and try all links to confirm navigation.


vrothberg:
One thing I would find useful is to copy the path of a specific disk image.  Once created, I may change to the terminal to copy/move it around or upload it somewhere.  Is it possible to have such a copy-path function?

Screenshot 2024-10-17 at 14 40 08

cdrage (approved):
Threw me off for a second thinking a lot of code was changed, but it was because EmptyScreen was deleted and re-created.

Heavily tested this out with POC of https://github.com/containers/podman-desktop-extension-bootc/issues/895

Code LGTM.

vrothberg (approved):
> Either way, none of this is directly related to this issue 🙂 so I've opened https://github.com/containers/podman-desktop-extension-bootc/issues/935, maybe we can continue the discussion there, or you can open a new issue for anything I've missed.

Let's do this.  Thank you!

Great work, LGTM

Issue #864, opened 2024-09-27
Share examples to help onboard new users


deboer-tim:
### Epic domain

Make it as easy as possible for users to create bootable containers, starting with a curated list of examples or "recipes". For instance, examples for building images for databases, a virt host, etc.

There is already a repository that we may be able to leverage for this: https://gitlab.com/fedora/bootc/examples

castrojo:
I was just trying to build a custom image this evening and have some ideas! The httpd, tailscale, and other httpd examples are pretty "appdeveloper cloud person" oriented. I was thinking from the other direction, highlighting the operating system since that's a killer feature in bootc. I was thinking something like grouping them into use cases:

- CentOS9 server
- CentOS9 Kubernetes Node
- CentOS10 Instruclab Workshop 
- CentOS9 Home Server  
- Fedora rPI Digital Sign
- ... and so on
 
- Fedora Silverblue desktop
- Fedora Kinoite desktop
- HeliumOS Workstation
- Bluefin Workstation
- ... and so on 

And then you FROM the usual barebones image, but then recommend some groups to install that match those use cases, and then leave a little boilerplate for "Add your packages here!". Then these examples automatically become useful to homelabbers and enthusiasts. And then if appropriate add images from orgs in ADOPTERS.md, and maybe having those orgs in the examples directory will incentivize them to submit their logo. 😄

cdrage:
> I was just trying to build a custom image this evening and have some ideas! The httpd, tailscale, and other httpd examples are pretty "appdeveloper cloud person" oriented. I was thinking from the other direction, highlighting the operating system since that's a killer feature in bootc. I was thinking something like grouping them into use cases:
> 
> ```
> - CentOS9 server
> - CentOS9 Kubernetes Node
> - CentOS10 Instruclab Workshop 
> - CentOS9 Home Server  
> - Fedora rPI Digital Sign
> - ... and so on
>  
> - Fedora Silverblue desktop
> - Fedora Kinoite desktop
> - HeliumOS Workstation
> - Bluefin Workstation
> - ... and so on 
> ```
> 
> And then you FROM the usual barebones image, but then recommend some groups to install that match those use cases, and then leave a little boilerplate for "Add your packages here!". Then these examples automatically become useful to homelabbers and enthusiasts. And then if appropriate add images from orgs in ADOPTERS.md, and maybe having those orgs in the examples directory will incentivize them to submit their logo. 😄

That would be awesome and would love to add that.

[...]

PR #942, opened 2024-10-17
feat: add examples page

cdrage:
feat: add examples page

### What does this PR do?

* Minor edit to category (see json update)
* Adds an example page (rendered portion) that shows all the examples we
  currently have
* Pull and build image buttons
* Reactively do in-progress for pulling

### Screenshot / video of UI

https://github.com/user-attachments/assets/9a9d67a3-ecc8-4367-aea9-1061d2fa81de

Screenshot 2024-10-17 at 3 34 15 PM
Screenshot 2024-10-17 at 3 34 25 PM

Closes https://github.com/containers/podman-desktop-extension-bootc/issues/895

### How to test this PR?

1. Go to examples page
2. Press pull image button
3. Then build image button after it finishes


cdrage:
Tests have been added! Ready for testing out and using 🚀 

I have added some "follow up issues" to the topic here, especially lack of examples (only 3 "good" ones right now): https://github.com/containers/podman-desktop-extension-bootc/issues/864#issuecomment-2420381794

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+  // Update the available bootc images if we receive a msg image pull update from the UI
+  return rpcBrowser.subscribe(Messages.MSG_IMAGE_PULL_UPDATE, async msg => {
+    bootcAvailableImages = await bootcClient.listBootcImages();
+
+    if (msg.image) {
+      const example = examples.find(example => example.image === msg.image);
+      if (example) {
+        example.state = 'pulled';
+        console.log(examples);
```
benoitf, line 26: to be removed

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
...
+        examples = [...examples];
+      }
+    }
+  });
+});
+
+// Function to update examples based on available images
+function updateExamplesWithPulledImages() {
+  if (bootcAvailableImages) {
+    bootcAvailableImages.forEach(image => {
```
benoitf, line 36: I think the linter is now recommending. for of pattern

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
...
+      // Only do it if there is a RepoTags
+      if (image.RepoTags) {
+        const [imageRepo, imageTag] = image.RepoTags[0].split(':');
```
benoitf, line 39: maybe use the optional ? pattern to reduce the if if if chaining

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
...
+          example.state = 'pulled';
+          examples = [...examples];
+        }
+      }
+    });
+  }
+}
+
+// Reactive statement to call the function each time bootcAvailableImages updates
+$: if (bootcAvailableImages) {
```
benoitf, line 52: might be a time to use $effect / svelte 5 runes

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
...
+  updateExamplesWithPulledImages();
+}
+</script>
+
+<Card title={category.name} classes="{$$props.class} font-medium mt-4">
+  <div slot="content" class="w-full">
+    {#if examples.length === 0}
+      <div class="text-gray-400 mt-2">There is no example in this category.</div>
```
benoitf, line 60: should ot use empty screen ?

packages/frontend/src/lib/ExamplesCard.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="ts">
+import Card from './Card.svelte';
+import ExampleCard from './ExampleCard.svelte';
+import type { Example, Category } from '../../../shared/src/models/examples';
```
benoitf, line 4: might replace all the ../.. by `@something/`

packages/frontend/src/lib/ExampleCard.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+import { render, fireEvent, screen } from '@testing-library/svelte';
+import { expect, test, vi } from 'vitest';
+import ExampleCard from './ExampleCard.svelte';
+import { bootcClient } from '../api/client';
```
benoitf, line 23: use absolute import

packages/frontend/src/lib/Card.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
```
benoitf, line 1: might switch to $props / runes

packages/frontend/src/Examples.svelte
```diff
@@ -0,0 +1,56 @@
...
+let groups: Map<Category, Example[]> = new Map();
+
+const UNCLASSIFIED: Category = {
+  id: 'unclassified',
+  name: 'Unclassified',
+};
+
+onMount(async () => {
+  // onmount get the examples
+  let examples = await bootcClient.getExamples();
```
benoitf, line 17: it means that it's only when we create the component, it won't be reactive it's ok ?

packages/backend/assets/examples.json
```diff
@@ -38,7 +38,7 @@
   "categories": [
     {
       "id": "fedora",
-      "name": "Fedora BootC Images"
+      "name": "Fedora Bootable Container Images"
```
benoitf, line 41: could be a separate small PR

deboer-tim (commented):
Ditto to all the points @benoitf raised.

Worked fine when testing. The biggest gap to me was how to find details or instructions for each example. Why does it exist, what does it do, and how do I use it? The source goes to the repo root, then I have to guess at the folder and look at the readme. (and ok, Tailscale doesn't have anything)

Could we link directly to the folder, or ideally have a way to view the readme directly? Maybe needs a details page for each one?

Noted a few other minor usability issues:
- The page appears, _then_ we switch buttons from Pull to Build. Is the API fast enough that we could do this before opening the page, or can we delay adding the button until we know which it is? (might need to move Source button to Link on the left so it doesn't move).
- Like on the dashboard, you don't know the image you're about to pull is huge. Would be nice to include the size on the card, or warn the user when they click on the button.
- No pull task or progress. Unrelated to this issue so I opened up #944.

cdrage:
> Ditto to all the points @benoitf raised.
> 
> Worked fine when testing. The biggest gap to me was how to find details or instructions for each example. Why does it exist, what does it do, and how do I use it? The source goes to the repo root, then I have to guess at the folder and look at the readme. (and ok, Tailscale doesn't have anything)
> 
> Could we link directly to the folder, or ideally have a way to view the readme directly? Maybe needs a details page for each one?
> 
> Noted a few other minor usability issues:
> 
> * The page appears, _then_ we switch buttons from Pull to Build. Is the API fast enough that we could do this before opening the page, or can we delay adding the button until we know which it is? (might need to move Source button to Link on the left so it doesn't move).
> * Like on the dashboard, you don't know the image you're about to pull is huge. Would be nice to include the size on the card, or warn the user when they click on the button.
> * No pull task or progress. Unrelated to this issue so I opened up [Usability of pulling images #944](https://github.com/containers/podman-desktop-extension-bootc/issues/944).

We have https://gitlab.com/fedora/bootc/examples/-/issues/19 which I could work on next sprint with regards to descriptions of what each one does.

We have no way of knowing the size of the image before pulling, but we can at least add an approximate "size" on the description when pulling?

ACK on button delay / showing it correctly.

packages/backend/assets/examples.json
cdrage: ACK

packages/frontend/src/lib/ExamplesCard.svelte
cdrage: Can't as we use `$$props.class` in the Card (copied from AI lab and we try to keep consistent with them too) which we can then not use since it puts the svelte file in "runes mode"

packages/frontend/src/lib/ExamplesCard.svelte
cdrage: No need :) will always have examples! 

Tried your suggestion though using empty screen but messed up formatting, so may be easier to just use this.

packages/frontend/src/Examples.svelte
cdrage: okay to not be reactive as we as not making any changes to the assets file reactively / modifying it / or getting it from a local file.

packages/frontend/src/lib/ExamplesCard.svelte
cdrage: NVM, we don't actually use classes imported into it, so it's okay to switch to svelte 5 runes, updating now.

cdrage:
@deboer-tim @benoitf

Updated the code based on the review:

- Added size of the image (hardcoded in examples.json), makes it easier and less calls to check for size (honestly sizes will not change much, and this is just for showing examples)
- Modified how we show the buttons on load, less flickering / changes.
- Did not update to svelte5 (could not get it to work very well without causing extra changes in other svelte files)
- Updated rest of minor code edits

cdrage:
> I'm no longer getting Pull buttons when I don't have the image locally: Screenshot 2024-10-21 at 5 03 30 PM

that's my mistake, I had messed up an `if` statement.

Latest commit I have also updated the styling, adding a "querying" button in case it takes a while to get the list of images, etc, as well as reduce the amount of flickering on initial load.

packages/frontend/src/lib/Card.svelte
```diff
@@ -0,0 +1,33 @@
...
+export let title: string | undefined = undefined;
+export let description: string | undefined = undefined;
+export let classes: string = '';
+
+export let href: string | undefined = undefined;
+</script>
+
+<a class="no-underline" href={href}>
+  <div class="{classes} rounded-md flex-nowrap overflow-hidden" role="region" aria-label={title ?? 'Card'}>
```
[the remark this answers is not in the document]
deboer-tim: We are doing this in a few places, often with different parameter names. In Svelte 5 you can do:
```
let {
  class: className = '',
}: {
  class?: string;
} = $props();
```
to make this cleaner, but since we ever only ever do `classes="font-medium mt-4"`, maybe remove the parameter and put the styling in here?

packages/frontend/src/lib/ExampleCards.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+import { render, screen } from '@testing-library/svelte';
+import { expect, test, vi } from 'vitest';
+import type { Example, Category } from '/@shared/src/models/examples';
+import ExamplesCard from './ExamplesCard.svelte';
```
deboer-tim, line 23: component: ExamplesCard, test: ExampleCards. Can you rename one so that they match?

packages/frontend/src/Examples.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
+import type { Example, Category } from '../../shared/src/models/examples';
```
deboer-tim, line 2: Should really use `/@shared/...` here, and other places.

deboer-tim (commented):
Thanks for the changes, worked well in testing. We have some ideas for additional improvements open, but this delivers on the core requirement.

Some minor requests for change but otherwise this is good to go for me.

deboer-tim (approved):
Thanks for the changes @cdrage, a nice little addition we can continue to expand on.

packages/frontend/src/Examples.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+  // Confirm example 2 is rendered
+  const example2 = screen.getByLabelText('Example 2');
+  expect(example2).toBeInTheDocument();
+
+  // Use the tick function to wait for the next render (updating pulled state)
+  await tick();
+
+  // Wait until example1 says Build image as it updates reactively
+  // same for example 2 but Pull image
+  waitFor(() => {
```
benoitf, line 131: issue: missing await before waitFor

packages/frontend/src/Examples.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    const buildImage1 = example1.querySelector('[aria-label="Build image"]');
+    expect(buildImage1).toBeInTheDocument();
+  });
+
+  waitFor(() => {
```
benoitf, line 136: issue: missing await before waitFor

packages/frontend/src/lib/Card.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+export let title: string | undefined = undefined;
+export let description: string | undefined = undefined;
+
+export let href: string | undefined = undefined;
```
benoitf, line 5: suggestion: use svelte5 runes

```suggestion
interface Props {
title: string;
description: string;
href: string
}

let {title, description, href} : Props = $props();
```

packages/frontend/src/lib/Card.svelte
```diff
@@ -0,0 +1,32 @@
...
+</script>
+
+<a class="no-underline" href={href}>
```
benoitf, line 8:
```suggestion
<a class="no-underline" {href}>
```

packages/frontend/src/lib/ExampleCard.svelte
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Function that takes something in MB and returns it in GB is more than 1GB
+// in a nice string format
+function formatStringSize(size: number): string {
+  if (size > 1000) {
+    return `${(size / 1000).toFixed(1)} GB`;
+  }
+  return `${size} MB`;
+}
```
benoitf, line 37: todo: please use humansize library

https://www.npmjs.com/package/human-size

packages/frontend/src/lib/ExampleCard.svelte
```diff
@@ -0,0 +1,111 @@
...
+
+// Make sure if example.pulled is updated, we force a re-render
+$: {
+  example.state;
+}
```
benoitf, line 42: issue: feels hacky

use svelte5 ($props, $effect, etc)

packages/frontend/src/lib/ExampleCard.svelte
```diff
@@ -0,0 +1,111 @@
...
+      {#if example?.state === 'pulled'}
+        <Button on:click={gotoBuild} icon={DiskImageIcon} aria-label="Build image" title="Build image" class="w-28"
+          >Build image</Button>
+      {:else if example?.state === 'unpulled'}
+        <Button
+          on:click={pullImage}
+          icon={faArrowDown}
+          aria-label="Pull image"
+          title="Pull image"
+          class="w-28"
```
benoitf, line 102: question: do we need to set a size there ?

packages/frontend/src/lib/ExampleCard.svelte
```diff
@@ -0,0 +1,111 @@
...
+          inProgress={pullInProgress}>Pull image</Button>
+      {:else}
+        <!-- Show a spinner / in progress for querying button instead if we are still loading information-->
+        <Button icon={faArrowDown} aria-label="Querying" title="Querying" class="w-28" inProgress={true}
```
benoitf, line 106: question: same here, why do we need to set a size there ? (w-28)

packages/frontend/src/lib/ExamplesCard.svelte
benoitf: I don't see any reason why it would fail

packages/frontend/src/lib/ExamplesCard.svelte
benoitf: todo: if we always have examples, then do not add the if clause

packages/frontend/src/lib/ExampleCard.svelte
cdrage: I decided to use the filesize library https://www.npmjs.com/package/filesize which is used in podman desktop. the one you linked hasn't been updated in 10 years.

cdrage:
Updated @benoitf @benoitf 

The more I use svelte5, the more I'm liking it! Much cleaner code.

I have updated it to svelte5 paradigms. I also decided to use https://www.npmjs.com/package/filesize library instead of humansize since it's the same on in Podman Desktop + is more updated.

packages/frontend/src/lib/ExampleCard.svelte
benoitf: ah yes I think I took the wrong name but you figured it out
I mixed humanize-duration and this one

benoitf:
yes the library name was wrongly spelled !

benoitf (approved):

cdrage:
> looks like there are some typechecking issue but approving

Thanks! Formatter had messed up and did not put `{}` around href. Added and updated now.

PR #1000, opened 2024-11-07
chore: refit terminal after initial loading to correct size

cdrage:
chore: refit terminal after initial loading to correct size

### What does this PR do?

`xterm.js` has an odd issue where it will not resize the terminal on
initial loading.

After a long time troubleshooting, there is no ideal fix as we are
loading the terminal within many other divs and it's unable to properly
refit itself.

This is causing issues with the scroll-bar being inaccurate. Even after
adding proper CSS / tailwind values for spacing, it is inaccurate / not
correct.

This adds a reactive statement that refits the terminal after a very
short 10ms delay after logs appear.

Closes https://github.com/podman-desktop/podman-desktop-extension-bootc/issues/997

### How to test this PR?

1. Load logs
2. No longer see padding on the bottom

### DIFF
--- a/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.svelte
+++ b/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.svelte
@@ -24,6 +24,16 @@ let termFit: FitAddon;
 let logsTerminal: Terminal;
 let logInterval: NodeJS.Timeout;
 
+// This is an issue with xterm not resizing properly when first initializing / loading
+// the terminal due to how we add padding. We must therefore call fit() approx 10ms after
+// initial loading to make sure the terminal is properly sized.
+// this is only called once when switching from no logs to logs
+$: if (noLogs === false) {
+  setTimeout(() => {
+    termFit?.fit();
+  }, 10);
+}
+
 async function fetchFolderLogs() {
   if (!folder) {
     return;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.svelte
```diff
@@ -24,6 +24,16 @@ let termFit: FitAddon;
...
 
+// This is an issue with xterm not resizing properly when first initializing / loading
+// the terminal due to how we add padding. We must therefore call fit() approx 10ms after
+// initial loading to make sure the terminal is properly sized.
+// this is only called once when switching from no logs to logs
+$: if (noLogs === false) {
+  setTimeout(() => {
+    termFit?.fit();
+  }, 10);
+}
```
benoitf, line 35: it's really odd to use a timeout ?

but anyway if I go to another component, timeout will occur so, the timeout needs to be cleared in onDestroy

packages/frontend/src/lib/disk-image/DiskImageDetailsVirtualMachine.svelte
```diff
@@ -35,6 +35,16 @@ let notifySubscriber: Subscriber;
...
+// This is an issue with xterm not resizing properly when first initializing / loading
+// the terminal due to how we add padding. We must therefore call fit() approx 10ms after
+// initial loading to make sure the terminal is properly sized.
+// this is only called once when switching from no logs to logs
+$: if (noLogs === false) {
+  setTimeout(() => {
+    termFit?.fit();
+  }, 10);
+}
+
```
benoitf, line 47: same remark here

it makes me feel that if you copy/paste the same exact content inside 2 files, probably you can just use one component rather than duplicating the changes (so any fixes will help the other side rather than duplicating the changes (and at some point missing the enhancement/fix in one part)

cdrage:
Closing this PR as this happens upstream too on Podman Desktop side.

PR #1026, opened 2024-11-18
feat: add build config configurator

cdrage:
feat: add build config configurator

### What does this PR do?

* Adds automated way to create a build config json file
* Added to build page

### Screenshot / video of UI

Screenshot 2024-11-18 at 3 44 47 PM

https://github.com/user-attachments/assets/bbf18809-620f-46f5-ad5e-a02c8df69fd0

Closes https://github.com/podman-desktop/extension-bootc/issues/827

### How to test this PR?

Use build config and configure / create your own custom build config
   with users, filesystem, etc.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -81,6 +81,11 @@ function toggleBuildConfig() {
   showBuildConfig = !showBuildConfig;
 }
 
+let showBuildConfigFile = false;
+function toggleBuildConfigFile() {
+  showBuildConfigFile = !showBuildConfigFile;
+}
+
 function findImage(repoTag: string): ImageInfo | undefined {
   return bootcAvailableImages.find(
     image => image.RepoTags && image.RepoTags.length > 0 && image.RepoTags[0] === repoTag,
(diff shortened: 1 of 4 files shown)


cdrage:
Discussed with @mrguitar on Slack, follow up issues after this PR:

https://github.com/podman-desktop/extension-bootc/issues/1027 and https://github.com/podman-desktop/extension-bootc/issues/1028

packages/frontend/src/Build.svelte
```diff
@@ -703,15 +882,22 @@ $: if (availableArchitectures) {
                     <Button on:click={() => getBuildConfigFile()}>Browse...</Button>
                   </div>
                   <p class="text-sm text-[var(--pd-content-text)] pt-2">
-                    The build configuration file is a TOML or JSON file that contains the build options for the disk
-                    image. Customizations include user, password, SSH keys and kickstart files. More information can be
-                    found in the <Link
-                      externalRef="https://github.com/osbuild/bootc-image-builder?tab=readme-ov-file#-build-config"
-                      >bootc-image-builder documentation</Link
-                    >.
+                    This will override any above user-specific input and use the supplied file only.
```
deboer-tim, line 885: The text at the top only talks about filling in the fields, then at the bottom of the form you find out you can also provide a file. Feels like maybe there should be two radio buttons like 'Create new build config' and 'Supply existing build config file' so that it's clear you have either option from the start?

packages/frontend/src/Build.svelte
```diff
@@ -684,14 +762,115 @@ $: if (availableArchitectures) {
...
-                on:click={toggleAdvanced}
-                ><Fa icon={showAdvanced ? faCaretDown : faCaretRight} class="inline-block mr-1" />Advanced Options
+                aria-label="build-config-options"
+                on:click={toggleBuildConfig}
+                ><Fa icon={showBuildConfig ? faCaretDown : faCaretRight} class="inline-block mr-1" />Build config
               </span>
-              {#if showAdvanced}
+              {#if showBuildConfig}
+                <p class="text-sm text-[var(--pd-content-text)] mb-2">
+                  Supplying the following fields will create a build config file that contains the build options for the
```
deboer-tim, line 771: It never says _where_ the build config will be generated. I think this would help (e.g. say we put it in the folder above), so that I could use this once to create it, then maybe edit it outside of Podman Desktop, then supply the same file next time I build.

deboer-tim (commented):
Very nice improvement 👍🏼, worked well in testing but just had some usability suggestions.

cdrage:
@deboer-tim @axel7083 

I've updated the PR with the following:
- Config.json is not created in the same directory as the output folder, so you can further edit it, etc.
- Updated the UI a bit so file and interactive mode are in different sections (we can improve it post-this-PR @deboer-tim for radio mode after I sync up with @MariaLeonova / UI/UX team)

Screenshot 2024-12-16 at 11 23 12 AM

benoitf (approved):

PR #1196, opened 2025-01-16
feat: Adds option to pull arm64/amd64 examples

cdrage:
feat: Adds option to pull arm64/amd64 examples

### What does this PR do?

* Adds the ability to choose what initial arm64 or amd64 image example
  you want
* Defaults to native architecture when initially loading
* Pulls the architecture correctly

### Screenshot / video of UI

https://github.com/user-attachments/assets/87effef0-74f4-4e4e-8131-0a5cf92a7928

Closes https://github.com/podman-desktop/extension-bootc/issues/1154

### How to test this PR?

1. Go to examples page
2. Select from the drop down the arch you want
3. Pull the image & build

### DIFF
--- a/packages/backend/assets/examples.json
+++ b/packages/backend/assets/examples.json
@@ -9,7 +9,7 @@
       "image": "registry.gitlab.com/fedora/bootc/examples/httpd",
       "tag": "latest",
       "categories": ["fedora"],
-      "architectures": ["amd64"],
+      "architectures": ["amd64", "arm64"],
       "basedir": "httpd",
       "size": 2000000000,
       "readme": "# HTTPD\n\nThis example provides an Apache HTTPD server exposed on port 80.\n\n## Building this example\n\n1. Build the image with either `podman build` or [Podman Desktop](https://podman-desktop.io/).\n2. (Optional for login and SSH access) Create a [\"config.toml\" or \"config.json\" build config](https://docs.fedoraproject.org/en-US/bootc/authentication/#_bootc_image_builder) that contains login and SSH information.\n3. Build the image with either [bootc-image-builder](https://github.com/osbuild/bootc-image-builder) or the [Podman Desktop BootC extension](https://github.com/containers/podman-desktop-extension-bootc) and (optionally) the \"config.toml\" you created.\n\n## Using this example\n\n1. Launch the virtual machine.\n2. Visit the VM IP address `http://<ip-address>` on your browser."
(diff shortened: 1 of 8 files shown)


deboer-tim (changes requested):
Need to update the copyright in the .ts files (`2024-2025`).

Instead of adding isArm() to the API, could we do getArch()? Seems a bit cleaner for future use. (or we could use os.arch() in the frontend?)

cdrage:
> Need to update the copyright in the .ts files (`2024-2025`).
> 
> Instead of adding isArm() to the API, could we do getArch()? Seems a bit cleaner for future use. (or we could use os.arch() in the frontend?)

We don't have getArch in the PD API, so instead I've added isArm and used the already-added isX86.

At the moment the only other arch that we could possibly support is x390 but I'm unsure what the status of that is.

I've updated the code / tests!

deboer-tim:
> > Need to update the copyright in the .ts files (`2024-2025`).

Thanks for updating!

> > Instead of adding isArm() to the API, could we do getArch()? Seems a bit cleaner for future use. (or we could use os.arch() in the frontend?)
> 
> We don't have getArch in the PD API, so instead I've added isArm and used the already-added isX86.
> 
> At the moment the only other arch that we could possibly support is x390 but I'm unsure what the status of that is.

But why not just expose a getArch() function directly, that just returns the value from os.arch [*]? It would be one less function and avoid potentially having to add something like isS390() in the future.

[*] I'm also finding it a bit odd that the UI uses isX86() to decide to default to the 'amd64' arch. One option would be to use the values from os.arch() consistently, but I think it would be cleaner for bootc.ts to include something like `export type Architecture = 'arm64' | 'amd64'`

Then the API could export one `getArch(): Architecture ` function, and both backend & shared pullImage() functions could have some type safety with `arch?: Architecture ` instead of `arch?: string`. Sound ok?

cdrage:
> > > Need to update the copyright in the .ts files (`2024-2025`).
> 
> Thanks for updating!
> 
> > > Instead of adding isArm() to the API, could we do getArch()? Seems a bit cleaner for future use. (or we could use os.arch() in the frontend?)
> > 
> > 
> > We don't have getArch in the PD API, so instead I've added isArm and used the already-added isX86.
> > At the moment the only other arch that we could possibly support is x390 but I'm unsure what the status of that is.
> 
> But why not just expose a getArch() function directly, that just returns the value from os.arch [*]? It would be one less function and avoid potentially having to add something like isS390() in the future.
> 
> [*] I'm also finding it a bit odd that the UI uses isX86() to decide to default to the 'amd64' arch. One option would be to use the values from os.arch() consistently, but I think it would be cleaner for bootc.ts to include something like `export type Architecture = 'arm64' | 'amd64'`
> 
> Then the API could export one `getArch(): Architecture ` function, and both backend & shared pullImage() functions could have some type safety with `arch?: Architecture ` instead of `arch?: string`. Sound ok?

Sounds good! I'll update it for that. For some reason I thought you meant adding "getArch" upstream to PD, but I see we don't have anything up there.

I'll add the getArch function and you are right, instead we can just do if / switch statements to determine if something is arm64 / amd64, etc.

cdrage:
Updated to use getArch instead (in a separate commit) and ready for a re-review! @deboer-tim

deboer-tim (approved):
Thanks Charlie, this is better!

As per my last comment the one thing I still don't like is how arch is just a string and the frontend needs to understand how to map that string between os.platform arch and container arch. I'd prefer to have BootcAPI getArch() and pullImage() use the same values (probably getArch() does some mapping) so that the UI can use the values directly and never needs code like this:
```
if (hostArch === 'x64') {
    defaultArch = 'amd64';
```
but I won't hold this up, we can take it as a follow-up.

cdrage:
> Thanks Charlie, this is better!
> 
> As per my last comment the one thing I still don't like is how arch is just a string and the frontend needs to understand how to map that string between os.platform arch and container arch. I'd prefer to have BootcAPI getArch() and pullImage() use the same values (probably getArch() does some mapping) so that the UI can use the values directly and never needs code like this:
> 
> ```
> if (hostArch === 'x64') {
>     defaultArch = 'amd64';
> ```
> 
> but I won't hold this up, we can take it as a follow-up.

I agree. I'll follow this up in a separate PR.

PR #1224, opened 2025-01-23
chore: use env for windows and linux

### LINKED ISSUE #1207: Use Podman Desktop env API for all platforms

machine-utils.ts was updated to start using the Podman Desktop env API for Mac, but Linux and Windows use os.platform instead.

Use Podman Desktop env api for all platforms.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Just updates windows and linux to use PD env API instead of os.platform(), so that it is consistent with mac. Tests mock to mac for now, as they're not doing anything dependent on execution OS.

### Screenshot / video of UI

N/A

Fixes #1207.

### How to test this PR?

PR checks.

### DIFF
--- a/packages/backend/src/machine-utils.spec.ts
+++ b/packages/backend/src/machine-utils.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


packages/backend/src/machine-utils.spec.ts
```diff
@@ -51,6 +51,11 @@ vi.mock('@podman-desktop/api', async () => {
         };
       },
     },
+    env: {
+      isMac: true,
```
benoitf, line 55: it seems it'll test only macos case in the unit tests ?

Usually it's a vi.fn function and there are tests for each os

deboer-tim: There are no tests that are actually platform dependent (yet), but agree, that's better for now and more obvious later. Updated.

cdrage (approved):

benoitf (approved):

PR #1238, opened 2025-01-28
chore: update to tailwindcss v4

benoitf:
### What does this PR do?
Update to tailwindcss v4

note:
- postcss config is no longer, need to install a separate package. But now vite plugin is provided so switch to this plug-in instead
- part to include in the css is no longer using the same stuff `@tailwind` ---> `@import`
- the config file is no longer included by default as well... https://tailwindcss.com/docs/upgrade-guide#using-a-javascript-config-file
- some rules have been renamed (like rounded or shadow or flex-grow) https://tailwindcss.com/docs/upgrade-guide#changes-from-v3

related to https://github.com/podman-desktop/podman-desktop/issues/10795

### How to test this PR?

please carefully review all the screen !!!


axel7083:
pnpm-lock.yaml is conflicting

benoitf:
> pnpm-lock.yaml is conflicting

yes but it's like egg and chicken problem. If I rebase but nobody test it I'll do rebase all the time while fixing the lock is only a small task.

axel7083 (approved):
LGTM

benoitf:
rebased (and now it included v4.0.1 as a new version is out)

PR #1249, opened 2025-01-29
fix: update setting nav test to match 1.16

### LINKED ISSUE #1235: adapt code to the changes in the ui package of podman-desktop

see https://github.com/podman-desktop/extension-bootc/pull/1228
it seems some changes in the library is producing some regression (unit tests)

fix the code or then fix in upstream to be able to switch to a more recent version of the ui package

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Tests are checking that when you're at each route, the corresponding nav item (and only that) is selected in the navigation. SettingsNavItem added an extra div in the hierarchy so test s checking the wrong div.

Ideally SettingsNavItem would have aria labels, and the test would look at more of the hierarchy - but it was only a minute to find the problem, fix the test, and add the missing test for examples. If it happens again we can do something better but for now just update the test (and add the missing paths).

### Screenshot / video of UI

N/A

Fixes #1235.

### How to test this PR?

PR test passes.

### DIFF
--- a/packages/frontend/src/Navigation.spec.ts
+++ b/packages/frontend/src/Navigation.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


gastoner (commented):
Codewise LGTM
just fix the linter check

benoitf (approved):

PR #1277, opened 2025-02-05
chore(test): revert pw framework update

cbr7:
### What does this PR do?
Tries to failing e2e tests that started not working 2 days ago.

https://github.com/podman-desktop/extension-bootc/issues/1278


cbr7:
@deboer-tim looks like dependobot updated the PW framework to version 1.15.1 two days ago, at which point the e2e tests started failing, I'm not sure what is causing the failures, since the update seems to be working fine elsewhere, but in the meantime I made this PR to revert the update and get the tests running while we investigate.

cbr7:
PR check seems to be passing now but running e2e tests on windows still has the same issue, so I don't think this PR fixed whatever problem there is, at least not completely.

deboer-tim (approved):
It's up to you guys. I'm ok reverting if it gives more chance to investigate.

In general I'd prefer moving to a place where e2e tests must pass and are testing only the current branch. Nightly wouldn't work for that either (since it's testing the previous one, or could fail if we're fixing a previous release like 1.6.1?), but again, it's better than before and could improve later.

PR #1285, opened 2025-02-05
chore: fix bootc label colouring on dark / light mode

cdrage:
chore: fix bootc label colouring on dark / light mode

### What does this PR do?

Fixes the colour showing in dark / light mode on the image page. As
before it was not picking up the tailwind-specific colour.

### Screenshot / video of UI

Screenshot 2025-02-05 at 3 21 18 PM

https://github.com/user-attachments/assets/9a8e3953-5eda-4377-ad98-4cd51d8d69d4

Closes https://github.com/podman-desktop/extension-bootc/issues/1174

### How to test this PR?

1.Pull a bootc test image
2. See that the label shows now

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -76,8 +76,8 @@
           "badge": {
             "label": "bootc",
             "color": {
-              "dark": "bg-sky-300",
-              "light": "bg-sky-300"
+              "dark": "#7dd3fc",
+              "light": "#7dd3fc"
             }
           }
         }


cdrage:
> FWIW I had tried and we could use some of the PD variables like: `"dark": "bg-[var(--pd-badge-builtin-extension-bg)]"` but others don't work (couldn't determine a reason). Either way, I couldn't find any that were both logically correct and a useful/similar color, so I agree this is the best fix for now.

I tried a few other ways too to try and "use" the colour coding too. I agree, let's fix this for now

PR #1313, opened 2025-02-10
feat: add anaconda iso module enable / disable

cdrage:
feat: add anaconda iso module enable / disable

### What does this PR do?

* Adds the ability to disable or enable any anaconda iso modules
* Added to build page with ability to choose enable/disable and
  add/remove as many as you want.

### Screenshot / video of UI

https://github.com/user-attachments/assets/e20ae912-3c81-49ed-8095-9ff6f69aa477

Screenshot 2025-02-10 at 12 19 38 PM

Closes https://github.com/podman-desktop/extension-bootc/issues/1028

### How to test this PR?

1. Add a few entries from within the build page.
2. Check your `config.json` after creation (it's in the same folder as
   your output folder).
3. Confirm modules have been put in there.

### DIFF
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -561,3 +561,24 @@ test('test building with a buildConfig JSON file that a temporary file for build
     expect(options.HostConfig.Binds[2]).toContain('config.json:ro');
   }
 });
+
+test('expect createBuildConfigJSON to work with anaconda iso modules being enabled / dsiabled', async () => {
+  const buildConfig = {
+    anacondaIsoInstallerModules: {
+      enable: ['test-module', 'test-module2'],
+      disable: ['test-module3', 'test-module4'],
+    },
+  } as BuildConfig;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const buildConfigJson: Record<string, any> = createBuildConfigJSON(buildConfig);
+  expect(buildConfigJson).toBeDefined();
+
+  // Expect enable to contain test-module and 2
+  expect(buildConfigJson.customizations.installer.modules.enable).toContain('test-module');
+  expect(buildConfigJson.customizations.installer.modules.enable).toContain('test-module2');
+
+  // Expect disable to contain test-module3 and 4
+  expect(buildConfigJson.customizations.installer.modules.disable).toContain('test-module3');
+  expect(buildConfigJson.customizations.installer.modules.disable).toContain('test-module4');
+});
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -562,7 +562,7 @@ test('test building with a buildConfig JSON file that a temporary file for build
   }
 });
 
-test('expect createBuildConfigJSON to work with anaconda iso modules being enabled / dsiabled', async () => {
+test('expect createBuildConfigJSON to work with anaconda iso modules being enabled / disabled', async () => {
   const buildConfig = {
     anacondaIsoInstallerModules: {
       enable: ['test-module', 'test-module2'],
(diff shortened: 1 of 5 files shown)


packages/shared/src/models/bootc.ts
```diff
@@ -26,6 +26,7 @@ export interface BuildConfig {
   user?: BuildConfigUser[];
   filesystem?: BuildConfigFilesystem[];
   kernel?: BuildConfigKernel;
+  anacondaIsoInstallerModules?: BuildConfigAnacondaIsoInstallerModules[];
```
deboer-tim, line 29: It feels a little like the API is written to support the current UI as opposed to being functional. How about something like this, and letting the frontend build the lists instead of backend?
```
anacondaIsoInstallerModulesEnable?: string[];
anacondaIsoInstallerModulesDisable?: string[];
```

deboer-tim (changes requested):
Functionally worked great in testing. 👍🏼

I don't love the UI though. Initially I was going to suggest using a checkbox instead of Dropdown for enable/disable, but I think it still gets complicated when you have multiple modules and the enable and disable are interleaved.

The underlying config is just 'list of strings to enable' and 'list of strings to disable'. Could we just have two columns (enable and disable), and each side is either a textarea (i.e. each separate line is a module) or set of Inputs?

cdrage:
@deboer-tim Updated the UI, great suggestion regarding the columns! Looks a lot better.

Screenshot 2025-02-11 at 1 03 06 PM

I decided to keep `anacondaIsoInstallerModules` as one though, but handle the enable/disable login within the renderer instead. 

I'm fine either way for separating them, but let me know if the changes look good enough that we don't need to change it up / have multiple anacondaIsoInstallerModules variables

deboer-tim (commented):
UI is much better now - thank you! 🎉.

I still think the API should be based on string arrays to match the config, either my previous suggestion or just:
```
port interface BuildConfigAnacondaIsoInstallerModules {
  enable: string[];
  disable: string[];
}
```
We already have separate arrays for `enabledAnacondaModules / disabledAnacondaModules` which could just be string[] and passed in directly. Should be a little cleaner Build.svelte without the `, enable: true` objects, and build-disk-image would be a little simpler too.

packages/backend/src/build-disk-image.spec.ts
```diff
@@ -561,3 +561,24 @@ test('test building with a buildConfig JSON file that a temporary file for build
     expect(options.HostConfig.Binds[2]).toContain('config.json:ro');
   }
 });
+
+test('expect createBuildConfigJSON to work with anaconda iso modules being enabled / dsiabled', async () => {
```
deboer-tim, line 565: Typo: dsiabled

deboer-tim (approved):
Looks good, tested well, API & code simpler. Thank you for the changes!

One nit: I think it removed blank entries before and now it doesn't, would be nice if build-disk-image filtered.

packages/frontend/src/Build.svelte
```diff
@@ -881,6 +913,56 @@ $: if (availableArchitectures) {
                     id="buildConfigKernelArguments"
                     placeholder="Kernel arguments (ex. quiet)"
                     class="w-full" />
+
+                  <!-- The below section is related to anaconda-iso, only show if buildType has 'anaconda-iso' in it -->
```
feloy, line 917: Is this comment still relevant? I don't understand how it is implemented

feloy (approved):

packages/frontend/src/Build.svelte
cdrage: Woops! Leftover comment should of been removed. Good catch.

cdrage:
> Looks good, tested well, API & code simpler. Thank you for the changes!
> 
> One nit: I think it removed blank entries before and now it doesn't, would be nice if build-disk-image filtered.

Thanks! Had to refactor the "removeEmptyStrings" function to get this to work so that there is no more blank sections. I've added this to a separate commit.

cdrage:
Tests pass! Been LGTM'd multiple times. Going to merge this :)

PR #1314, opened 2025-02-10
chore: using simple objects in rpc api (to avoid snapshots)

### LINKED ISSUE #1304: Virtual Machine tab no longer refreshing
### Bug description

Similar to: https://github.com/podman-desktop/extension-bootc/issues/1303

Virtual Machine tab never updates past loading...

Image

Image

### Operating system

mac M1

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Moving to svelte 5, state objects are proxied. Normally that's an implementation detail, but when you try to pass the proxies through rpc they can't be cloned and result in errors.

We could use $state.snapshot(build) every time we pass objects from the UI to the API, but we really shouldn't be passing whole objects to the backend and blindly trusting that the objects (including all properties) match what the backend knows anyway.

This changes the three affected API functions to accept ids instead of objects, and the backend looks up the objects before acting. Although it's slightly more work on the backend, this is a better/more robust API, and also avoids the svelte 5 proxy issue.

### Screenshot / video of UI

N/A

Fixes #1304.

### How to test this PR?

PR checks, delete a build, start a VM. Tests added to check we're passing strings.

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -109,7 +109,7 @@ export class BootcApiImpl implements BootcApi {
       // create an array of builds. invalid build ids are ignored
       const builds = buildIds
         .map(id => this.history.getHistory().find(build => build.id === id))
-        .filter(build => build) as BootcBuildInfo[];
+        .filter(build => !!build);
 
       // Map each build to a delete operation promise
       const deletePromises = builds.map(build => this.deleteBuildContainer(build));
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -57,7 +57,7 @@ export class BootcApiImpl implements BootcApi {
   async checkVMLaunchPrereqs(buildId: string): Promise<string | undefined> {
     const build = this.history.getHistory().find(build => build.id === buildId);
     if (!build) {
-      return Promise.reject(new Error(`Could not find build: ${buildId}`));
+      throw new Error(`Could not find build: ${buildId}`);
     }
     return createVMManager(build).checkVMLaunchPrereqs();
   }
@@ -344,7 +344,6 @@ export class BootcApiImpl implements BootcApi {
   // this method is internal and meant to be used by the API implementation
   protected async notify(id: string, body: unknown = {}): Promise<void> {
     // Must pass in an empty body, if it is undefined this fails
-    console.log('notify frontend: ' + id);
     await this.webview.postMessage({
       id,
       body,
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -54,7 +54,11 @@ export class BootcApiImpl implements BootcApi {
     return checkPrereqs(await getContainerEngine());
   }
 
-  async checkVMLaunchPrereqs(build: BootcBuildInfo): Promise<string | undefined> {
+  async checkVMLaunchPrereqs(buildId: string): Promise<string | undefined> {
+    const build = this.history.getHistory().find(build => build.id === buildId);
+    if (!build) {
+      return Promise.reject(new Error(`Could not find build: ${buildId}`));
+    }
     return createVMManager(build).checkVMLaunchPrereqs();
   }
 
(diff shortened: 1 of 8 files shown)


packages/backend/src/api-impl.ts
```diff
@@ -54,7 +54,11 @@ export class BootcApiImpl implements BootcApi {
     return checkPrereqs(await getContainerEngine());
   }
 
-  async checkVMLaunchPrereqs(build: BootcBuildInfo): Promise<string | undefined> {
+  async checkVMLaunchPrereqs(buildId: string): Promise<string | undefined> {
+    const build = this.history.getHistory().find(build => build.id === buildId);
+    if (!build) {
+      return Promise.reject(new Error(`Could not find build: ${buildId}`));
```
benoitf, line 60: question : why promise.rejectand not throw new Error ?

packages/backend/src/api-impl.ts
```diff
@@ -330,6 +344,7 @@ export class BootcApiImpl implements BootcApi {
   // this method is internal and meant to be used by the API implementation
   protected async notify(id: string, body: unknown = {}): Promise<void> {
     // Must pass in an empty body, if it is undefined this fails
+    console.log('notify frontend: ' + id);
```
benoitf, line 347: remove ?

packages/backend/src/api-impl.ts
deboer-tim: Should be throw error, fixed in new commit.

packages/backend/src/api-impl.ts
deboer-tim: Removed.

packages/backend/src/api-impl.ts
```diff
@@ -90,13 +99,18 @@ export class BootcApiImpl implements BootcApi {
...
     const response = await podmanDesktopApi.window.showWarningMessage(
       `Are you sure you want to remove the selected disk images from the build history? This will remove the history of the build as well as remove any lingering build containers.`,
       'Yes',
       'No',
     );
     if (response === 'Yes') {
+      // create an array of builds. invalid build ids are ignored
+      const builds = buildIds
+        .map(id => this.history.getHistory().find(build => build.id === id))
+        .filter(build => build) as BootcBuildInfo[];
```
benoitf, line 112: issue: since TypeScript 5.5 / https://github.com/microsoft/TypeScript/pull/57465 the inferring is automatic on the filter so no need to add a cast to as `BootcBuildInfo[]`

but use `filter(build => !!build)` rather than `(build => build)`

https://www.typescriptlang.org/play/?ts=5.7.3#code/C4TwDgpgBAsiDyAjAVhAxsKBeKBvAlgCYBcUAdgK4C2iEATgDRTAQAewpAzsHfmQOYBfANwAoUWgD2ZblCoIU6YJ1JwkqDAG0AutiiaCJKAEYmLdqQDkAM0mTLgpodIAmM2w5RLiAIZ0HTkSkpsweVr4AXg7aYhLSsvicMD7AaAAWfPzqSnoAFJIoqgoawACU2AB8eKJQUPjWUPkoAHRE2Fg4xuW4NbVQdBDAFHRkUAXIvYJQEAA2nNA9ff2Dw6MUZIQQ1nwQhGK1U5NQsVIymNszLAOEevLZGJzNVD5gTciVdUkp6Zn3wG+lUpiU6yMjSACqGy2O0IABVZhAqBAyMo9Bcrrtmuj6Ll8CwqB8AISEvGIoFxM5QHwzGYAMUk60IKjUii0uhwYLIkM22zIu3hM0RyOUJ3ikkFzRmkn4uUsYPIEF2zEkVMINw8dB8UDQPm4TAGnAol0+liY1LpDI2nFKQA

deboer-tim: Nice! I wasn't aware of that, fixed.

benoitf (approved):
thx

cdrage (approved):
Functionality fixes the VM page now, thank you!

Implementation is great

Issue #975, opened 2024-10-28
Color, component tech debt


deboer-tim:
### Is your enhancement related to a problem? Please describe

I noticed some minor cleanup that should be done in Tailwind config and Build page.

### Describe the solution you'd like

- Tailwind config has a lot of hard-coded colors, many are already unused and can be removed.
- `<select>` control in the Build page can now be a Dropdown from the UI library.
- `<a>` control on the Build page should be a Link or changed.
- Once these are done, more colors can be removed from Tailwind config.

cdrage:
Also to add:

Specifically these sections too regarding "dropdown": https://github.com/podman-desktop/extension-bootc/blob/f86033d656ae4bb001ae98007249cd57a33c821d/packages/frontend/src/Build.svelte#L912

cdrage:
I see a lot of commits cleaning up the hardcoded colours.

I believe we can close this now as we now rely on PD upstream colour library rather than our hardcoded ones.

PR #1337, opened 2025-02-14
chore: svelte 5 for build, link, markdown

deboer-tim:
### What does this PR do?

Moves some of the core components like Dashboard, Link, and Markdown to Svelte 5. Removed aria-label prop since it is unused.

### Screenshot / video of UI

N/A

Part of #975.

### How to test this PR?

Check UI to make sure no change in behaviour.


packages/frontend/src/lib/Link.svelte
```diff
@@ -19,6 +24,6 @@ async function click(): Promise<void> {
 }
 </script>
 
-<Link title={title} aria-label={$$props['aria-label']} on:click={click}>
-  <slot />
+<Link title={title} on:click={click}>
```
benoitf, line 27: question: where the ariaLabel property went ?

packages/frontend/src/lib/dashboard/Dashboard.svelte
```diff
@@ -66,11 +66,11 @@ onMount(async () => {
 });
 
 // Each time bootcAvailableImages updates, check if 'quay.io/bootc-extension/httpd' is in RepoTags
-$: {
+$effect(() => {
   if (bootcAvailableImages?.some(image => image.RepoTags?.includes(exampleImage))) {
     imageExists = true;
   }
-}
+});
```
benoitf, line 73: imageExists should be a derived state rather than having an effect updating the imageExists state

packages/frontend/src/lib/Link.svelte
deboer-tim: See description - it was totally unused, so I figured there was no point keeping or porting it and easy enough to add back later. If you'd like, I can keep it.

packages/frontend/src/lib/dashboard/Dashboard.svelte
deboer-tim: Good catch, fixed.

benoitf (approved):

packages/frontend/src/lib/Link.svelte
benoitf: yes, it's only that I prefer to have cleanup being a separate operation of migrating a component

cleanup = removing
migrate = translate but no addition/removal

deboer-tim:
Rebased (no merge conflicts, but GitHub doesn't want to merge old branch).

PR #1338, opened 2025-02-14
chore: format vite config

deboer-tim:
### What does this PR do?

Just formats vite config files.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

N/A

### How to test this PR?

N/A

PR #1340, opened 2025-02-16
Switch to biome for formatting

deboer-tim:
### What does this PR do?

From a comment in #1338 - we're missing formatting of config files.

If we're going to fix formatting, I figured we may as well copy from Podman Desktop and upgrade to / align with biome. First commit copies in the biome config and package.json changes, second commit is the first pass running it.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

PR check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   "scripts": {
     "build": "concurrently \"cd packages/frontend && npm run build\" \"cd packages/backend && npm run build\"",
     "watch": "concurrently \"cd packages/frontend && npm run watch\" \"cd packages/backend && npm run watch\"",
-    "format:check": "prettier --check \"**/src/**/*.{ts,svelte}\"",
-    "format:fix": "prettier --write \"**/src/**/*.{ts,svelte}\"",
+    "format:check": "biome format",
+    "format:fix": "biome format --write",
     "lint:check": "eslint . --cache",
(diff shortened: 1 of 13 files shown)


benoitf (approved):
FYI biomejs is not formatting the html part of svelte files

if you're ok keeping them unformatted then you're free to go

package.json
```diff
@@ -13,8 +13,8 @@
   "scripts": {
     "build": "concurrently \"cd packages/frontend && npm run build\" \"cd packages/backend && npm run build\"",
     "watch": "concurrently \"cd packages/frontend && npm run watch\" \"cd packages/backend && npm run watch\"",
-    "format:check": "prettier --check \"**/src/**/*.{ts,svelte}\"",
-    "format:fix": "prettier --write \"**/src/**/*.{ts,svelte}\"",
+    "format:check": "biome format",
```
benoitf, line 16: markdown files won't be formatted/checked

benoitf:
note: pr's title is not compliant with conventional commits

feloy (approved):

package.json
deboer-tim: Yes, that's unfortunately. I will likely put up another PR to start using prettier for md the same as podman-desktop repo. (checked the biome roadmap, they're hoping to add md this year but it is well down the list)

PR #1339, opened 2025-02-14
chore: remove a href from build page

deboer-tim:
### What does this PR do?

The build page still had a link using hard-coded color values, just convert to a Link like all the others.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Part of #975.

### How to test this PR?

When you have no bootc images the link should still appear and work.

PR #1346, opened 2025-02-18
chore: replace spy with vi.mocked

deboer-tim:
### What does this PR do?

From a comment on #1339, we use a spy in several places where we should now be using vi.mocked(). This just cleans up the low hanging fruit.

### Screenshot / video of UI

N/A

Part of #975.

### How to test this PR?

Code review, PR checks still green.

### DIFF
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -41,6 +41,8 @@ const config: Configuration = {
   update: vi.fn(),
 };
 
+vi.mock('node:fs');
+
 vi.mock('@podman-desktop/api', async () => {
   return {
     env: {
@@ -250,7 +252,6 @@ test('check build exists', async () => {
 
   // mock two existing builds on disk: qcow2 and vmdk
   const existsList: string[] = [resolve(folder, 'qcow2/disk.qcow2'), resolve(folder, 'vmdk/disk.vmdk')];
-  vi.mock('node:fs');
   vi.spyOn(fs, 'existsSync').mockImplementation(f => {
     return existsList.includes(f.toString());
   });
@@ -540,8 +541,6 @@ test('test building with a buildConfig JSON file that a temporary file for build
     folder: '/foo/bar/qemutest4',
     buildConfig: buildConfig,
   } as BootcBuildInfo;
-
-  vi.mock('node:fs');
 
   const options = createBuilderImageOptions(name, build);
 
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -541,9 +541,7 @@ test('test building with a buildConfig JSON file that a temporary file for build
     buildConfig: buildConfig,
   } as BootcBuildInfo;
 
-  // Spy on fs.writeFileSync to make sure it is called
   vi.mock('node:fs');
-  vi.spyOn(fs, 'writeFileSync');
 
   const options = createBuilderImageOptions(name, build);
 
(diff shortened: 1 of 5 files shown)


packages/frontend/src/Build.spec.ts
```diff
@@ -548,7 +545,7 @@ test('Show the image if isManifest: true and Labels is empty', async () => {
   render(Build, { imageName: 'testmanifest1', imageTag: 'latest' });
 
   await vi.waitFor(() => {
-    expect(spyOnInspectManifest).toHaveBeenCalled();
+    expect(vi.mocked(bootcClient.inspectManifest)).toHaveBeenCalled();
```
feloy, line 548: I think you can simplify with:
```
expect(bootcClient.inspectManifest).toHaveBeenCalled();
```

packages/backend/src/build-disk-image.spec.ts
```diff
@@ -541,9 +541,7 @@ test('test building with a buildConfig JSON file that a temporary file for build
     buildConfig: buildConfig,
   } as BootcBuildInfo;
 
-  // Spy on fs.writeFileSync to make sure it is called
   vi.mock('node:fs');
```
axel7083, line 544: Shouldn't this be at the root of the file ?

packages/frontend/src/Build.spec.ts
```diff
@@ -669,7 +663,7 @@ test('have amd64 and arm64 NOT disabled if inspectManifest contains both archite
...
```
axel7083, line 666:
```suggestion
    expect(bootcClient.inspectManifest).toHaveBeenCalled();
```

Can be simplified: vite is very nice for helping us and directly consider a mock when we pass it to expect, it would throw an error if the object is not an mock

packages/frontend/src/lib/disk-image/DiskImageActions.spec.ts
```diff
@@ -65,15 +65,11 @@ test('Test clicking on delete button', async () => {
...
-  // spy on deleteBuild function
-  const spyOnDelete = vi.spyOn(bootcClient, 'deleteBuilds');
-
   // Click on delete button
   const deleteButton = screen.getAllByRole('button', { name: 'Delete Build' })[0];
   deleteButton.click();
 
-  expect(spyOnDelete).toHaveBeenCalled();
-  expect(spyOnDelete).toHaveBeenCalledWith(['name1']);
+  expect(vi.mocked(bootcClient.deleteBuilds)).toHaveBeenCalledWith(['name1']);
```
axel7083, line 72:
```suggestion
  expect(bootcClient.deleteBuilds).toHaveBeenCalledWith(['name1']);
```

packages/frontend/src/Build.spec.ts
deboer-tim: I was trying to clean up one bad pattern and copied another one 😬. Fixed these in new commit, including all the other instances in this repo.

packages/backend/src/build-disk-image.spec.ts
deboer-tim: I'm assuming we should also be using vi.mocked() over vi.spyOn()? I was just trying to fix one problem at a time, but I did move this mock() to root in a new commit.

packages/frontend/src/Build.spec.ts
deboer-tim: 👍🏼 Fixed.

packages/frontend/src/lib/disk-image/DiskImageActions.spec.ts
deboer-tim: Fixed.

axel7083 (approved):
Thanks for the cleanup ! 🚀

PR #1355, opened 2025-02-19
fix: correctly kick off build in background

### LINKED ISSUE #1302: Build always ends in timeout
### Bug description

Most likely related to our change of timeout to 5 seconds maybe? 

Each time I go build I always end up with a timeout now:

https://github.com/user-attachments/assets/7c22ebbf-f34f-4619-95dc-22b3884083c8

### Operating system

mac m1 / latest build

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

cc @deboer-tim

deboer-tim:
### What does this PR do?

Either during increasing our linting rules and properly handling promises, or updating the message proxy, the behaviour of returning a promise from buildDiskImage() changed. Instead of returning the window.withProgress promise immediately, it now waits until the task/promise is complete before returning. As a result, the build always 'times out' and you receive an error in the UI.

The correct behaviour should be returning immediately after starting the background task. This just adds a catch block so that we can remove the return and not await.

I'll be honest, I have no idea how to correctly test for this behaviour. :)

### Screenshot / video of UI

N/A

Fixes #1302.

### How to test this PR?

See #1302. Start a build and confirm you're redirected to the build logs instead of receiving a timeout error.

### DIFF
--- a/packages/backend/src/build-disk-image.spec.ts
+++ b/packages/backend/src/build-disk-image.spec.ts
@@ -25,13 +25,17 @@ import {
   getBuilder,
   getUnusedName,
   createBuildConfigJSON,
+  buildDiskImage,
 } from './build-disk-image';
 import { bootcImageBuilderCentos, bootcImageBuilderRHEL } from './constants';
 import type { ContainerInfo, Configuration } from '@podman-desktop/api';
+import * as extensionApi from '@podman-desktop/api';
 import { containerEngine } from '@podman-desktop/api';
 import type { BootcBuildInfo, BuildConfig } from '/@shared/src/models/bootc';
 import * as fs from 'node:fs';
+import type { History } from './history';
 import path, { resolve } from 'node:path';
+import { getContainerEngine } from './container-utils';
 
 const configurationGetConfigurationMock = vi.fn();
 
@@ -54,11 +58,22 @@ vi.mock('@podman-desktop/api', async () => {
     configuration: {
       getConfiguration: (): Configuration => config,
     },
+    window: {
+      withProgress: vi.fn(),
+      showErrorMessage: vi.fn(),
+    },
+    ProgressLocation: {
+      TASK_WIDGET: 'TASK_WIDGET',
+    },
   };
 });
 
+vi.mock('./container-utils');
+
+vi.mock('./machine-utils');
+
 beforeEach(() => {
-  vi.clearAllMocks();
+  vi.resetAllMocks();
 });
 
 test('check image builder options', async () => {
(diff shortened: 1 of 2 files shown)


benoitf (commented):
how to test:


window.withProgress = long task like 2s wait

assert that await of the function completes before a call after the withProgress wait of 2s

should fail with current code, works with the PR code

benoitf (approved):

packages/backend/src/build-disk-image.spec.ts
```diff
@@ -54,6 +57,27 @@ vi.mock('@podman-desktop/api', async () => {
...
+      TASK_WIDGET: 'TASK_WIDGET',
+    },
+  };
+});
+
+vi.mock('./container-utils', async () => {
+  return {
+    getContainerEngine: vi.fn().mockReturnValue({
+      connection: { type: 'podman', status: vi.fn().mockReturnValue('started') },
+    } as unknown as ProviderContainerConnection),
```
axel7083, line 74: The mockReturnValue should be done in the beforeEach, running `vi.resetAllMock()` remove all the mockReturnValue.

packages/backend/src/build-disk-image.spec.ts
```diff
@@ -54,6 +57,27 @@ vi.mock('@podman-desktop/api', async () => {
...
+  };
+});
+
+vi.mock('./machine-utils', async () => {
+  return {
+    checkPrereqs: vi.fn().mockReturnValue(undefined),
```
axel7083, line 80: same remark

packages/backend/src/build-disk-image.spec.ts
axel7083: if the beforeEach(() => { vi.resetAllMock() }) is missing it should be added

packages/backend/src/build-disk-image.ts
```diff
@@ -112,9 +112,9 @@ export async function buildDiskImage(build: BootcBuildInfo, history: History, ov
   const telemetryData: Record<string, unknown> = {};
   telemetryData.build = build;
 
-  // "Returning" withProgress allows PD to handle the task in the background with building.
+  // Kick off task using withProgress to build in the background
   let errorMessage: string;
-  return extensionApi.window
+  extensionApi.window
```
axel7083, line 117: IMHO it does not make much sense to have a function with the following signature (promise)

https://github.com/podman-desktop/extension-bootc/blob/8fc7abafe8e4b5d5c29a439c43f41759a16be85a/packages/backend/src/build-disk-image.ts#L60

and having a dangling promise not awaited at the end. If the caller do not to bother to wait, this is the caller issue, not the method implementation. 

The signature of the method is misleading then. In ai-lab we solved it by having two method:

```
// request => do not wait for the task to be finishes, only start it (do not return a promise)
requestCreateInferenceServer(): string;

// awaiting this promise will wait for full completion (return a promise)
createInferenceServer(): Promise<string>;
```

packages/backend/src/build-disk-image.spec.ts
deboer-tim: Switched to using resetAllMock(), but since only the one new test needs getContainerEngine() to be mocked I moved the mocking into the test. (and found I don't actually need to mock machine-utils return value, just need basic mock()).

packages/backend/src/build-disk-image.spec.ts
deboer-tim: Fixed.

packages/backend/src/build-disk-image.ts
deboer-tim: Understood, but launching a build requires validation, checking prereqs on Podman, updating the build history, then launching the 2+ minute task to actually do the build. We don't want a function that waits for the full build to complete, but even those initial things require awaits and might take a couple seconds.

IMHO having the promise is accurate and still allows you to decide between awaiting to confirm that the build actually kicked off vs backgrounding it. Would a comment or renaming help with your concern instead?

deboer-tim:
Rebased to fix merge conflicts.

packages/backend/src/build-disk-image.ts
deboer-tim: @axel7083 ok? This PR is just 'fixing the current intended behaviour' so even if you think this is wrong I'd prefer to fix the bug and take a followup. :)

cdrage (approved):
Was able to test and it redirects fine, rather than erroring out.

I agree with @deboer-tim should follow up in separate PR as this will fix functionality before release.

PR #1286, opened 2025-02-06
chore: enable promises linting

deboer-tim:
### What does this PR do?

Enable the remaining two linting rules for promises, even redoing some of the code I 'fixed' last week because there's a better change once you have more linting enabled.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Fixes #1245.

### How to test this PR?

Code review; PR checks.

PR #1356, opened 2025-02-19
fix: refresh log timeout not triggering

### LINKED ISSUE #1303: Build log no longer refreshing / updating
### Bug description

On the page, build log is no longer updating even though we are writing to the file (can confirm in log file it's writing).

https://github.com/user-attachments/assets/b576ed9b-f072-41cd-b067-4b03a24da189

### Operating system

mac m1 / latest build

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

cc @deboer-tim

deboer-tim:
### What does this PR do?

#1286 caused a regression where fetchFolderLogs is no longer called every 2s. Added () and a catch so that it's called correctly. Also, refreshInterval is a constant so shouldn't be part of the state.

Tempted to make the interval a parameter to reduce test time, but started by just adding a test that waits to confirm the timeout starts triggering with 3.5s.

### Screenshot / video of UI

N/A

Fixes #1303.

### How to test this PR?

Start a build, open logs, and confirm that the output keeps updating.

### DIFF
--- a/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.spec.ts
+++ b/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.spec.ts
@@ -16,7 +16,7 @@
  ***********************************************************************/
 
 import { render, screen, waitFor } from '@testing-library/svelte';
-import { vi, test, expect, beforeAll } from 'vitest';
+import { vi, test, expect, beforeAll, beforeEach } from 'vitest';
 import DiskImageDetailsBuild from './DiskImageDetailsBuild.svelte';
 import { bootcClient } from '/@/api/client';
 import type { Subscriber } from '/@shared/src/messages/MessageProxy';
@@ -50,6 +50,10 @@ beforeAll(() => {
   });
 });
 
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
 const mockLogs = `Build log line 1
 Build log line 2
 Build log line 3`;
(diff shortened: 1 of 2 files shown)


gastoner (approved):
LGTM tested on Windows

packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.spec.ts
```diff
@@ -81,3 +85,21 @@ test('Handles empty logs correctly', async () => {
...
+
+  render(DiskImageDetailsBuild, { folder: '/empty/logs' });
+
+  // verify we start refreshing logs
+  await vi.waitFor(
+    () => {
+      expect(bootcClient.loadLogsFromFolder).toHaveBeenCalledTimes(2);
+    },
+    {
+      timeout: 3500,
```
benoitf, line 101:
```suggestion
      timeout: 3_500,
```

nit

packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.svelte
```diff
@@ -18,7 +18,7 @@ let { folder }: Props = $props();
 let logsXtermDiv = $state<HTMLDivElement>();
 let noLogs = $state(true);
 let previousLogs = $state('');
-const refreshInterval = $state(2000);
+const refreshInterval = 2000;
```
benoitf, line 21:
```suggestion
const refreshInterval = 2_000;
```
nit

benoitf (approved):

PR #1448, opened 2025-03-18
chore: allow dependencies to be built with pnpm v10

odockal:
### What does this PR do?
Explicitly state the dependencies with build scripts to execute during ie. install lifecycle phase in pnpm v10. This will enable to e2e tests to run.

We have alternatives, I went with the minimal working solution, other options:
```
"onlyBuiltDependencies": ["electron", "@biomejs/biome", "esbuild", "svelte-preprocess"] -> building those that appeared on the warning
OR
"neverBuiltDependencies": [] -> to build any
```
When I run `pnpm approve-builds` locally, it updated pnpm-workspace.yaml file instead of package.json.
https://github.com/podman-desktop/extension-bootc/issues/1447

### How to test this PR?
E2E tests should be passing now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,8 @@
   "pnpm": {
     "overrides": {
       "vite>esbuild": "^0.25.0"
-    }
+    },
+    "neverBuiltDependencies": []
   },
   "packageManager": "pnpm@9.10.0+sha512.73a29afa36a0d092ece5271de5177ecbf8318d454ecd701343131b8ebc0c1a91c487da46ab77c8e596d6acf1461e3594ced4becedf8921b074fbd8653ed7051c"
 }


benoitf (changes requested):
electron should not be a 'known' dependency from the extension pov

odockal:
Until it is resolved in the core [Refactor the type of dependency for electron tests-playwright package](https://github.com/podman-desktop/podman-desktop/issues/11859), we cannot do much in the PR now.

benoitf:
@odockal 

well you could add

If you want the old pre v10 behaviour, so you want to allow all dependencies to run postinstall scripts, then add this to your package.json:

```json
{
  "pnpm": {
    "neverBuiltDependencies": []
  }
}
```

and it'll be like pnpm 9.x

package.json
```diff
@@ -7,7 +7,7 @@
   "publisher": "redhat",
   "private": true,
   "engines": {
-    "node": ">=20.9.0",
+    "node": ">=22.0.0",
```
jeffmaury, line 10: Seems unrelated

benoitf (approved):

Issue #862, opened 2024-09-27
Table of bootable containers


deboer-tim:
### Is your enhancement related to a problem? Please describe

We show (built) disk images, but if you want to see your bootable containers you need to go to Images, where it's hard to see the list of bootable ton

### Describe the solution you'd like

Add a page similar to Images within the extension, but only show bootable containers.

In keeping with Podman Desktop design, we will likely need to create a subnav like Kubernetes or AI Lab and have multiple pages within the extension.

PR #1481, opened 2025-04-01
feat: bootable images icon

deboer-tim:
### What does this PR do?

When I added the resource card for bootc images to the Dashboard I reused the 'disk image' icon because we didn't have one for bootc images. I forgot we actually do have an icon that's used for bootc images on the Podman Desktop Images page.

This just turns the same icon (from the woff in backend) into a Svelte component so that it can be used by the frontend, and switches the Dashboard to use it.

### Screenshot / video of UI

Bootc images in the Images page (just showing which icon is currently used):

Screenshot 2025-03-31 at 9 26 33 PM

Dashboard before:

Screenshot 2025-03-31 at 9 24 51 PM

Dashboard after:

Screenshot 2025-03-31 at 5 05 15 PM

Icon needed as part of #862, noticed the discrepancy on the dashboard while creating it.

### How to test this PR?

Confirm image change on the Dashboard.


gastoner (commented):
For Icon LGTM
But the change with the `/images/` path is not working for me:

https://github.com/user-attachments/assets/ba92e0d8-77af-4604-9dd7-0ad8e2fab850

deboer-tim:
> For Icon LGTM But the change with the `/images/` path is not working for me:

Doh, sorry - that was part of another PR. Removed the link change, clicking on images should go nowhere before or after this change again (until we complete #862).

benoitf (approved):

cdrage (approved):
Appears and confirm it works.

LGTM!

PR #1487, opened 2025-04-01
chore: added progress verbose flag to image builder

### LINKED ISSUE #1305: Disable "progress bar" of bootc-image-builder container.

New bootc-image-builder introduces a "progress bar" that's enabled by default:

Image

This messes up our logging output / it's no longer very verbose.

We can disable this via: https://github.com/osbuild/bootc-image-builder?tab=readme-ov-file#detailed-description-of-optional-flags I believe with --progress=verbose to no longer use the progress bar.

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds `--progress verbose` flag to image builder

Closes #1305 

### How to test this PR?
Try to build an image, you should no longer see the progress bar. Instead, you should see detailed log


benoitf (approved):

feloy:
I can see in the issue that the prefered way would be to fix the terminal display instead. Did you try that direction? (by setting a different TERM for example?)

axel7083 (approved):
LGTM Codewise

gastoner:
> I can see in the issue that the prefered way would be to fix the terminal display instead. Did you try that direction? (by setting a different TERM for example?)

Not really since it is not goal of the issue

feloy (approved):

Issue #1488, opened 2025-04-02
`Pull image` button state for first time user does not persist when leaving the webview


axel7083:
### Bug description

When pulling the image on low internet connection this may take a while, but leaving the page and coming back, the button is restored to the default state.

https://github.com/user-attachments/assets/7fb06ea5-ae1b-4c77-9feb-55a390bcbe5b

Same for examples page

![Image](https://github.com/user-attachments/assets/9e2eac6d-df87-4248-8bc5-8058636f1f9e)

### Operating system

Windows 11

### Version

next (development version)

### Steps to reproduce

This is a bit weird, because I try to pull an example, and I don't see any task in the Task Manager, and the state lost if I leave the page

### Relevant log output

```shell

```

PR #1489, opened 2025-04-02
chore: refactor image pull logic and provide image store

deboer-tim:
### What does this PR do?

For disk images (historyInfo) we use a store that's updated via an event (MSG_HISTORY_UPDATE). For container images things are a lot more complicated:
- The list of images on the Dashboard and Examples pages was not reactive, just comes from the API on page load.
- pullImage() is async - but the promise would never fail.
- Since bootc images tend to be > 1Gb, pullImage() almost always times out in the RPC browser.
- A MSG_IMAGE_PULL_UPDATE event was fired and the pages would update state accordingly based on the event.
- The dashboard has a disclaimer after 5s - but this would show in 10s or not at all b/c pullImage would time out.

This commit fixes pullImage and replaces the event logic with an image store:
- pullImage() now returns errors if they occur, and is added to the 'no timeout' list so that it works async correctly.
- A generic MSG_IMAGE_UPDATE is fired when the pull is complete.
- A new ImageInfo store provides a reactive object identical to the disk image store.
- Pages are updated to use the store and remove event logic.

I don't particularly want to create another store, but they are not deprecated in Svelte 5 and using a store moves us in the right direction: consistency for both types of images, and pages updated to use a single/shared reactive object.

### Screenshot / video of UI

N/A

Required for #862.

### How to test this PR?

Delete existing bootc images if you have then.

Go to the Bootc Dashboard and pull the example image, make sure the button disables, disclaimer appears after 5s, and when the pull is complete the button re-enables, disclaimer disappears, and image count updates.

Go to the Examples page, pull an image and ensure the button disables, and re-enables/updates when complete.

Do not change pages while pulling, you'll hit unrelated issue #1488. :)

### DIFF
--- a/packages/frontend/src/lib/dashboard/Dashboard.svelte
+++ b/packages/frontend/src/lib/dashboard/Dashboard.svelte
@@ -2,16 +2,15 @@
 import BootcSelkie from '../BootcSelkie.svelte';
 import Link from '../Link.svelte';
 import { faArrowCircleDown, faCube } from '@fortawesome/free-solid-svg-icons';
-import { onMount, tick } from 'svelte';
-import { bootcClient, rpcBrowser } from '../../api/client';
-import { Messages } from '/@shared/src/messages/Messages';
 import { router } from 'tinro';
+import { tick } from 'svelte';
+import { bootcClient } from '../../api/client';
 import { Button, Expandable } from '@podman-desktop/ui-svelte';
-import type { ImageInfo } from '@podman-desktop/api';
 import DashboardPage from '../upstream/DashboardPage.svelte';
 import DashboardResourceCard from '../upstream/DashboardResourceCard.svelte';
 import DiskImageIcon from '../DiskImageIcon.svelte';
 import DashboardGuideCard from '../upstream/DashboardGuideCard.svelte';
+import { imageInfo } from '../../stores/imageInfo';
 import { historyInfo } from '../../stores/historyInfo';
 import osbuildImage from './osbuild.png';
 import redhatImage from './redhat.png';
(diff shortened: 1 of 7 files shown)


axel7083 (approved):
LGTM, tested following the steps you provided (including staying on the same page :p)

packages/frontend/src/lib/dashboard/Dashboard.svelte
```diff
@@ -52,36 +48,18 @@ async function pullExampleImage(): Promise<void> {
       // Ensure UI updates to reflect the new state
       tick().catch((e: unknown) => console.error('error updating disclaimer', e));
     }
-  }, 5000);
-}
+  }, 5_000);
 
-onMount(async () => {
-  bootcAvailableImages = await bootcClient.listBootcImages();
+  await bootcClient.pullImage(exampleImage);
```
axel7083, line 53: something more robust would to do the following
```suggestion
  await bootcClient.pullImage(exampleImage).finally(() => {
    pullInProgress = false;
    displayDisclaimer = false;
  });
```

this would ensure that if pullImage throw an error, we still stop the loading animation

axel7083: You can also use the non-promise version

```suggestion
  try {
    await bootcClient.pullImage(exampleImage);
  } finally {
    pullInProgress = false;
    displayDisclaimer = false;
  }
```

deboer-tim: You're right, thanks. 👍🏼 Went with the first option in new commit.

gastoner (approved):
LGTM on Linux

cdrage (approved):

PR #1493, opened 2025-04-03
chore(test): creating e2e tests for examples bootc page

cbr7:
### What does this PR do?
Adds new e2e tests to check building of example bootable images.

https://github.com/podman-desktop/extension-bootc/issues/1376

### DIFF
--- a/tests/playwright/src/bootc-extension.spec.ts
+++ b/tests/playwright/src/bootc-extension.spec.ts
@@ -211,7 +211,7 @@ test.describe('BootC Extension', () => {
 
   for (const example of examples) {
     test.describe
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):
Nice addition! Approving, one follow-up suggestion and a question:

- I'm not sure how to handle the image name mapping, but could we make this dynamic in the future? i.e. pull the list of examples from the page itself to avoid breakage when the list of examples is updated.
- Like all bootc images, the examples are large. When I manually run this earlier in the week it failed 1/4 times purely due to network. Do the e2e tests use a cache/proxy to help avoid stability issues / dependency on external registry?

PR #1516, opened 2025-04-08
chore(test): switching to smaller image to reduce test flakyness

cbr7:
### What does this PR do?
Tries to better handle test flakyness that comes from pulling big images on cloud environments.

### DIFF
--- a/tests/playwright/src/bootc-extension.spec.ts
+++ b/tests/playwright/src/bootc-extension.spec.ts
@@ -200,8 +200,8 @@ test.describe('BootC Extension', () => {
 
   const examples = [
     {
-      appName: 'Apache httpd',
-      imageName: 'registry.gitlab.com/fedora/bootc/examples/httpd:latest',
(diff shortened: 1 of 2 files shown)


axel7083 (commented):
Not sure the size reduction is significant?

- `registry.gitlab.com/fedora/bootc/examples/httpd:latest` is 1.7GB
- `registry.gitlab.com/fedora/bootc/examples/app-podman-systemd:latest` is 1.67

cbr7:
@axel7083 for me it shows the other as 1.61 GB, it's not much but it's something, I've been seeing the tests time out waiting for the image to be downloaded.

cbr7:
@axel7083 we're testing here the actual examples provided by the extension, I don't know if @deboer-tim would want to add some kind of fake image to use in testing, but even so we'd still need to check the real ones to make sure they work I'd think.

deboer-tim (approved):

PR #1524, opened 2025-04-09
chore: bootable images UI object

deboer-tim:
### What does this PR do?

As we add a bootable images page we need to convert images to a simpler UI object for easier display and manipulation as we've done in Podman Desktop. This is heavily copied from that code, with a bunch of simplification and cleanup for the bootc case.

Note: support for manifests is not included as it looks like the API is not exported from Podman Desktop, will be followup issue.

### Screenshot / video of UI

N/A

Part of #862.

### How to test this PR?

Just PR checks, code review.

### DIFF
--- a/packages/frontend/src/lib/images/ImageInfoUI.ts
+++ b/packages/frontend/src/lib/images/ImageInfoUI.ts
@@ -1,0 +1,30 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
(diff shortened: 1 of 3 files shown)


axel7083 (approved):

cdrage (approved):
Reviewed the code and thank you for the clean up, this helps with regards to the images.

If you can open up an issue for manifests (as the feature will no longer work), maybe we can iron that out before next release so we can no break any current features.

PR #1583, opened 2025-05-01
chore: update dependencies and CI workflows for macadam

cdrage:
chore: update dependencies and CI workflows for macadam

### What does this PR do?

* Update .github workflows
* Update dependencies in package.json and pnpm-lock.yaml

### Screenshot / video of UI

N/A

Reference: https://github.com/podman-desktop/extension-bootc/issues/1521

### How to test this PR?

Follow regular CI pipeline checks.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -41,6 +41,11 @@ jobs:
           node-version: 22
           cache: 'pnpm'
 
+      - name: add macadam package authentication
+        run: |
+          echo @crc-org:registry=https://npm.pkg.github.com >> ./.npmrc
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-check.yaml
```diff
@@ -118,6 +123,11 @@ jobs:
           run_install: false
           package_json_file: ./podman-desktop/package.json
 
+      - name: add macadam package authentication
```
deboer-tim, line 126: It's not clear to me why we need to add macadam auth here (as well as in the extension below)?

deboer-tim (approved):

.github/workflows/pr-check.yaml
cdrage: Ah you're right. Only need to add it to the extension below. I'll remove this one. 

I got confused since I saw "pnpm install" below it and thought it was the extension.

PR #1584, opened 2025-05-01
feat: add backend support for Macadam create image

cdrage:
feat: add backend support for Macadam create image

### What does this PR do?

* Adds Macadam create image backend support
* Adds unit tests and constants updates

### Screenshot / video of UI

N/A

Reference: https://github.com/podman-desktop/extension-bootc/issues/1521

### How to test this PR?

1. Run backend unit tests
2. Test API integration with frontend.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,8 +70,7 @@
     "typescript": "5.8.3",
     "typescript-eslint": "^8.31.1",
     "vite": "6.3.4",
-    "vitest": "^3.1.2",
-    "globals": "^16.0.0"
+    "vitest": "^3.1.2"
   },
   "dependencies": {
     "@xterm/addon-attach": "^0.11.0",
--- a/packages/backend/__mocks__/@podman-desktop/api.js
+++ b/packages/backend/__mocks__/@podman-desktop/api.js
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024-2025 Red Hat, Inc.
+ * Copyright (C) 2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,10 +15,4 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-
-// Image related
-export const bootcImageBuilder = 'bootc-image-builder';
-export const bootcImageBuilderCentos =
-  'quay.io/centos-bootc/bootc-image-builder:sha256-e53a3916cfc416f00a54a93757d7a48beb1af7fce3a3a329d07a0eea2e2b0737';
-export const bootcImageBuilderRHEL = 'registry.redhat.io/rhel9/bootc-image-builder:9.5';
-export const macadamName = 'rhel';
+export {};
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import * as podmanDesktopApi from '@podman-desktop/api';
+import type { CreateVmOptions, VmDetails } from '@crc-org/macadam.js';
 import type { ImageInfo, ContainerInfo } from '@podman-desktop/api';
 import type { BootcApi } from '/@shared/src/BootcAPI';
 import type { BootcBuildInfo, BuildType } from '/@shared/src/models/bootc';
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -70,8 +70,7 @@
     "typescript": "5.8.3",
     "typescript-eslint": "^8.31.1",
     "vite": "6.3.4",
-    "vitest": "^3.1.2",
-    "globals": "^16.0.0"
```
cdrage, line 74: Accidentally added 2 entries of "globals" and it merged in before I could update it! Hence change here.

packages/backend/__mocks__/@podman-desktop/api.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+export {};
```
cdrage, line 18: FYI @deboer-tim reason for this is because internally macadam.js also calls @podman-desktop/api.js and it was recommended to add a mocks functionality here to prevent conflicts.

Without it, it'll try and import @podman-desktop when referencing @crc-org/macadam.js within the tests.

packages/backend/src/api-impl.ts
```diff
@@ -70,6 +72,28 @@ export class BootcApiImpl implements BootcApi {
     return buildDiskImage(build, this.history, overwrite);
   }
 
+  // Launches a macadam VM by initializing the class and initializing the VM
+  async createVM(options: CreateVmOptions): Promise<void> {
+    const macadam = new MacadamVirtualMachine();
+    try {
+      await macadam.init();
```
deboer-tim, line 79: Is there any downside to creating a new MacadamVirtualMachine and initializing it every time? Or conversely does it get out to sync if you create it once and reuse later?

e.g. should we have a getMacadamVirtualMachine function that creates and inits it once, then returns the same instance the next time it is called? (doing this as follow-up is totally fine too)

packages/backend/src/macadam.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import * as macadam from '@crc-org/macadam.js';
+import { macadamName } from './constants';
+
+interface StderrError extends Error {
+  stderr?: string;
+}
+export class MacadamVirtualMachine {
```
deboer-tim, line 24: The naming threw me off above because this isn't a virtual machine, this is a wrapper or handler class for the macadam API which could list multiple VMs. Maybe rename to something like `MacadamHandler` or `MacadamSupport`?

packages/backend/src/macadam.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  const macadamVm = new MacadamVirtualMachine();
+
+  const macadamInstance = (macadam.Macadam as Mock).mock.results[0].value;
+  await macadamVm.init();
+
+  expect(macadam.Macadam).toHaveBeenCalledWith('rhel');
+  expect(macadamInstance.init).toHaveBeenCalled();
+});
+
+test('Test creating VM with MacadamVirtualMachine', async () => {
```
deboer-tim, line 48: Missing equivalent test for listVMs.

packages/backend/src/api-impl.ts
deboer-tim: After reviewing MacadamVirtualMachine I'm leaning more toward init() being an implementation details, i.e. line 79 should be removed and the class automatically calls init when you create it, or the first time you call createVM or listVM. (but I'm still wondering if we should only create one shared instance of MacadamVirtualMachine here)

deboer-tim (changes requested):
LGTM, just one missing test, a naming suggestion, and a question about sharing/refactoring a little.

packages/backend/src/api-impl.ts
cdrage: I've moved init() into the createVm and listVms.

I tried putting it into the constructor, but with our typescript lint checking, we do not allow it:
```
/Users/cdrage/syncthing/dev/electron/podman-desktop-extension-bootc/packages/backend/src/macadam.ts
  36:5  error  Refactor this asynchronous operation outside of the constructor  sonarjs/no-async-constructor
```

But I've at least moved init() into the functions so that we do not have to call init() each time in `api-impl.ts`.

deboer-tim (approved):
Still feels like we could reuse/share instances of macadam.Macadam instead of creating and initing each time, but good enough for now / easy to improve later.

Changes for the changes.

cdrage:
> Still feels like we could reuse/share instances of macadam.Macadam instead of creating and initing each time, but good enough for now / easy to improve later.
> 
> Thanks for the changes!

I agree! We can also look at whether `macadam.js` could improve this too / lazy load init() so we do not have to initialize it each time.

Issue #1588, opened 2025-05-02
Improve usability of creating VMs


deboer-tim:
### Is your enhancement related to a problem? Please describe

It works! But some ideas for additional UX features or improvements based on comments so far.

### Describe the solution you'd like

- Create VM button at top right of Disk Images page, beside the Build button.
- Change enablement of Create VM button on each disk image to disabled when the image type isn't compatible.
- Use a task when creating a VM, so that it's visible in the status bar / task manager.
- Provide a link to Resources.
- Review SSH parameters to see if there's anything else we can do to make it simpler or more obvious.

deboer-tim:
Fixes integrated for all of the listed issues. There is more work that can be done once we decide on the ultimate direction for Macadam integration, but in the short term this is complete.

PR #1589, opened 2025-05-02
chore: add create vm form to navigation api

deboer-tim:
### What does this PR do?

Adds the Create VM form to the navigation api, and uses it in the action on the disk images page. Adds navigation telemetry so we know how often users are going to this page.

Added tests, included missing ones for existing navigation.

### Screenshot / video of UI

N/A

Part of #1588.

### How to test this PR?

Confirm navigation still works to create VM form.


cdrage (approved):
LGTM and works well! No more having to use `router.goto` which is nice.

PR #1591, opened 2025-05-02
chore: add create vm button to disk image list

deboer-tim:
### What does this PR do?

Adds a Create VM button to the disk image list, matching what we have for building disk images (you can launch on an image to pre-fill the page, or you can just open the build page and select everything).

Since this is the first time the Create VM form is used without input params, the route has to be added and parameters made optional.

### Screenshot / video of UI

Screenshot 2025-05-02 at 2 55 26 PM

Part of #1588

### How to test this PR?

Open Disk Images page and confirm that the Create VM button is there and opens the form.

### DIFF
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -63,6 +63,9 @@ onMount(() => {
       <Route path="/disk-images/build/:name/:tag" breadcrumb="Build" let:meta>
         <Build imageName={decodeURIComponent(meta.params.name)} imageTag={decodeURIComponent(meta.params.tag)} />
       </Route>
+      <Route path="/disk-images/createVM" breadcrumb="Create Virtual Machine">
+        <CreateVM />
+      </Route>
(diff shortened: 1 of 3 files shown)


cdrage (changes requested):
Code LGTM, but I think we should have the "Build" button on the right, the Create VM on the left, to match Podman Desktop where we have the most "popular" buttons more on the far right.

Example:

Screenshot 2025-05-05 at 8 43 34 AM

cdrage (changes requested):
Code LGTM, but I think we should have the "Build" button on the right, the Create VM on the left, to match Podman Desktop where we have the most "popular" buttons more on the far right.

Example:

Screenshot 2025-05-05 at 8 43 34 AM

deboer-tim:
I was thinking about it left to right in 'order', but not tied to that - buttons switched.

cdrage (approved):
LGTM

PR #1594, opened 2025-05-05
chore: add more examples to the examples page

cdrage:
chore: add more examples to the examples page

### What does this PR do?

* Adds cloud init example, wifi and kernel module management examples.
* There are missing README's vs upstream for wifi and kernel module, so they have
been added here instead.

### Screenshot / video of UI

https://github.com/user-attachments/assets/2ab15249-4b8e-422d-88e0-927490a0af13

Closes https://github.com/podman-desktop/extension-bootc/issues/1510

### How to test this PR?

View examples page and see the new examples.


deboer-tim (commented):
Both the wifi and kernel module examples worked great: it shows both architectures and pulls down ARM64 by default on my Mac, which builds fine.

The cloud-init shows both architectures and appears to pull down ARM64, but when you go to build the image appears to be AMD64 and fails to build. Maybe this one is only AMD64?

Nit: the image sizes are all closer to 1.6Gb when I pull them.

jeffmaury (approved):
LGTM

deboer-tim:
@cdrage I don't think you pushed, as I still see cloud-init and not qemu-guest-agent.

cdrage:
> @cdrage I don't think you pushed, as I still see cloud-init and not qemu-guest-agent.

You're right, it didn't push! Should be there now.

feloy (approved):

Issue #1597, opened 2025-05-07
Macadam CLI and Resources integration


deboer-tim:
### Is your enhancement related to a problem? Please describe

Need to integrate the Settings > CLI tools and Resources for Macadam.

### Describe the solution you'd like

todo

deboer-tim:
CLI tools has no purpose at the minute so was not added, the rest of the support was added primarily via #1599, with a few PRs adding enablement and UI improvement in #1632. Work complete, closing.

PR #1599, opened 2025-05-07
feat: bootc vm provider

deboer-tim:
### What does this PR do?

Adds a Bootc VM provider to Settings > Resources. Only bootc extension created VMs appear in this section.

The bulk of code is direct copy pasta from the Macadam extension, which we'll need to keep in sync until we figure out how to share / where VMs belong. I've only added two 'eslint-disable sonarjs/no-ignored-exceptions' (better to keep code in sync than fix these in one place) and changed the labels/ descriptions.

Two new dependencies added: ssh2 and compare-versions.

### Screenshot / video of UI

Screenshot 2025-05-07 at 11 45 25 AM

Main part of #1597.

### How to test this PR?

Create a couple VMs, confirm they work identically to before (but without requiring another extension).

### DIFF
--- a/packages/backend/src/extension.ts
+++ b/packages/backend/src/extension.ts
@@ -26,8 +26,68 @@ import { HistoryNotifier } from './history/historyNotifier';
 import { Messages } from '/@shared/src/messages/Messages';
 import { satisfies, minVersion, coerce } from 'semver';
 import { engines } from '../package.json';
+import * as macadamJSPackage from '@crc-org/macadam.js';
+import { isHyperVEnabled, isWSLEnabled } from './macadam/win/utils';
+import { getErrorMessage, verifyContainerProivder } from './macadam/utils';
+import { LoggerDelegator } from './macadam/logger';
+import { ProviderConnectionShellAccessImpl } from './macadam/macadam-machine-stream';
+import { macadamName } from './constants';
 
 export const telemetryLogger = extensionApi.env.createTelemetryLogger();
+
+const EXTENSION_DESCRIPTION_MARKDOWN = `Bootable container-based virtual machines run via Macadam`;
+
+let stopLoop = false;
+
+type StatusHandler = (name: string, event: extensionApi.ProviderConnectionStatus) => void;
+const macadamMachinesInfo = new Map<string, MachineInfo>();
+const currentConnections = new Map<string, extensionApi.Disposable>();
+
+let wslAndHypervEnabledContextValue = false;
+const WSL_HYPERV_ENABLED_KEY = 'macadam.wslHypervEnabled';
+
+const listeners = new Set<StatusHandler>();
+
+export interface BinaryInfo {
+  path: string;
+  version: string;
+  installationSource: 'extension' | 'external';
+}
+
+export type MachineInfo = {
+  name: string;
+  image: string;
+  cpus: number;
+  memory: number;
+  diskSize: number;
+  port: number;
+  remoteUsername: string;
+  identityPath: string;
+  vmType: string;
+};
+
+type MachineJSON = {
+  Name: string;
+  Image: string;
+  CPUs: number;
+  Memory: string;
+  DiskSize: string;
+  Running: boolean;
+  Starting: boolean;
+  Port: number;
+  RemoteUsername: string;
+  IdentityPath: string;
+  VMType: string;
+};
+
+type MachineJSONListOutput = {
+  list: MachineJSON[];
+  error: string;
+};
+
+export let macadam: macadamJSPackage.Macadam;
+
+export const macadamMachinesStatuses = new Map<string, extensionApi.ProviderConnectionStatus>();
 
 export async function activate(extensionContext: ExtensionContext): Promise<void> {
   console.log('starting bootc extension');
@@ -114,6 +174,15 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
       await openBuildPage(panel, image);
     }),
   );
+
+  macadam = new macadamJSPackage.Macadam(macadamName);
+  await macadam.init();
+
+  const provider = await createProvider(extensionContext);
+
+  monitorMachines(provider, extensionContext).catch((error: unknown) => {
+    console.error('Error while monitoring machines', error);
+  });
 }
 
 function checkVersion(version: string): boolean {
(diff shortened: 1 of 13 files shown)


cdrage (changes requested):
Still testing the implementation, but so far so good, absolutely loving that it comes up on Resources now / no longer have to use the rhel vm extension.

Question though, could we perhaps have the code in `upstream` or something similar? 

So we can keep track on what's a direct copy-and-paste from the RHEL VM extension, and be able to update if there are any changes. So we don't end up in another "rpc proxy" code situation.

deboer-tim:
> Still testing the implementation, but so far so good, absolutely loving that it comes up on Resources now / no longer have to use the rhel vm extension.
> 
> Question though, could we perhaps have the code in `upstream` or something similar?
>
> So we can keep track on what's a direct copy-and-paste from the RHEL VM extension, and be able to update if there are any changes. So we don't end up in another "rpc proxy" code situation.

There's a bunch of code that has to be in extension.ts, or if we moved it out we'd have different code anyway, so I left that in. For everything else 'upstream' felt a little odd (since the issue is that there _isn't_ an upstream yet) so I went with 'macadam'.

PR rebased on latest and repushed.

packages/backend/src/extension.ts
```diff
@@ -26,9 +26,69 @@ import { HistoryNotifier } from './history/historyNotifier';
...
+const EXTENSION_DESCRIPTION_MARKDOWN = `Bootable container-based virtual machines run via Macadam`;
+
+let stopLoop = false;
+
+type StatusHandler = (name: string, event: extensionApi.ProviderConnectionStatus) => void;
+const macadamMachinesInfo = new Map<string, MachineInfo>();
+const currentConnections = new Map<string, extensionApi.Disposable>();
+
+let wslAndHypervEnabledContextValue = false;
+const WSL_HYPERV_ENABLED_KEY = 'macadam.wslHypervEnabled';
```
cdrage, line 47: should be in constants.ts file?

packages/backend/src/extension.ts
```diff
@@ -26,9 +26,69 @@ import { HistoryNotifier } from './history/historyNotifier';
...
 
 export const telemetryLogger = extensionApi.env.createTelemetryLogger();
 
+const EXTENSION_DESCRIPTION_MARKDOWN = `Bootable container-based virtual machines run via Macadam`;
+
+let stopLoop = false;
+
+type StatusHandler = (name: string, event: extensionApi.ProviderConnectionStatus) => void;
+const macadamMachinesInfo = new Map<string, MachineInfo>();
+const currentConnections = new Map<string, extensionApi.Disposable>();
```
cdrage, line 44: Unsure what the above do / no comments. I'm assuming that these are copypasta from the rhel vm extension?

packages/backend/src/extension.ts
```diff
@@ -26,9 +26,69 @@ import { HistoryNotifier } from './history/historyNotifier';
...
+import * as macadamJSPackage from '@crc-org/macadam.js';
+import { isHyperVEnabled, isWSLEnabled } from './macadam/win/utils';
+import { getErrorMessage, verifyContainerProivder } from './macadam/utils';
+import { LoggerDelegator } from './macadam/logger';
+import { ProviderConnectionShellAccessImpl } from './macadam/macadam-machine-stream';
+import { macadamName } from './constants';
 
 export const telemetryLogger = extensionApi.env.createTelemetryLogger();
 
+const EXTENSION_DESCRIPTION_MARKDOWN = `Bootable container-based virtual machines run via Macadam`;
```
cdrage, line 38: const file?

packages/backend/src/extension.ts
```diff
@@ -26,9 +26,69 @@ import { HistoryNotifier } from './history/historyNotifier';
...
+  Port: number;
+  RemoteUsername: string;
+  IdentityPath: string;
+  VMType: string;
+};
+
+type MachineJSONListOutput = {
+  list: MachineJSON[];
+  error: string;
+};
```
cdrage, line 86: Should the above inputs be in a separate file (types.ts) or something similar?

packages/backend/src/extension.ts
```diff
@@ -114,6 +174,15 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
       await openBuildPage(panel, image);
     }),
   );
+
+  macadam = new macadamJSPackage.Macadam(macadamName);
+  await macadam.init();
```
cdrage, line 179: Shouldn't we have if statements to only allow on Mac since that's what we support at the moment?

We should also have this in a try / catch since there may be issues initializing that would break monitorMachines as well as the general extension activation.

packages/backend/src/extension.ts
```diff
@@ -158,6 +227,340 @@ export async function getConfigurationValue<T>(property: string): Promise<T | un
   return extensionApi.configuration.getConfiguration('bootc').get<T>(property);
 }
 
+async function getJSONMachineList(): Promise<MachineJSONListOutput> {
```
cdrage, line 230: For getting json machine list, etc, should we have this instead in a separate file so it's more maintainable (since extension.ts is getting huge now).

packages/backend/src/extension.ts
```diff
@@ -158,6 +227,340 @@ export async function getConfigurationValue<T>(property: string): Promise<T | un
...
+    provider.updateStatus('stopped');
+  } catch (err: unknown) {
+    telemetryRecords.error = err;
+    throw err;
+  } finally {
+    // send telemetry event
+    const endTime = performance.now();
+    telemetryRecords.duration = endTime - startTime;
+    //in the POC we do not send any telemetry
+    //sendTelemetryRecords('macadam.machine.stop', telemetryRecords, false);
```
cdrage, line 344: surprised linter didn't catch, should we remove commented out code?

packages/backend/src/extension.ts
```diff
@@ -158,6 +227,340 @@ export async function getConfigurationValue<T>(property: string): Promise<T | un
...
+    }
+
+    const previousStatus = macadamMachinesStatuses.get(machine.Image);
+    if (previousStatus !== status) {
+      // notify status change
+      listeners.forEach(listener => listener(machine.Image, status));
+      macadamMachinesStatuses.set(machine.Image, status);
+    }
+
+    // TODO update cpu/memory/disk usage
```
cdrage, line 423: is this complete or in future PR

cdrage (changes requested):
Few comments regarding the extension.ts section.

for the `macadam/` folder, is that the files from the rhel vm extension? should we have  the tests copied over too?

cdrage:
> > Still testing the implementation, but so far so good, absolutely loving that it comes up on Resources now / no longer have to use the rhel vm extension.
> > Question though, could we perhaps have the code in `upstream` or something similar?
> > So we can keep track on what's a direct copy-and-paste from the RHEL VM extension, and be able to update if there are any changes. So we don't end up in another "rpc proxy" code situation.
> 
> There's a bunch of code that has to be in extension.ts, or if we moved it out we'd have different code anyway, so I left that in. For everything else 'upstream' felt a little odd (since the issue is that there _isn't_ an upstream yet) so I went with 'macadam'.
> 
> PR rebased on latest and repushed.

Did review before seeing this comment, but ACK to all of that.

cdrage:
Screenshot 2025-05-13 at 1 41 08 PM

Looks like it hasn't been updated yet / still says to install macadam extension

cdrage:
https://github.com/user-attachments/assets/47a57f1c-6c81-4b18-9cbc-e9c404709568

Unsure if this was fixed upstream or not, but still getting the issue of when you click start while on details and the terminal does not load correctly.

cdrage:
> I have a separate PR to clean up the messages in our frontend once this is merged, I was keeping this PR limited to only copying and integrating the macadam extension code.
> 
> Which is kind of the cause of all of your comments - at the point I copied this last week it was the macadam extension (now RHEL): there were no tests, all of the code in extensions.ts here was in extensions.ts there, and there are a couple instances of TODOs and commented-out code. It's not clear to me how much of this code we should clean up or refactor, since every change will make our code 'different', and hence harder to copy/paste any new updates from that extension.
> 
> So: I would prefer to do the UI changes in another PR, and keep this code as-is as much as possible so that we can copy in bug fixes or missing tests easily in subsequent PRs. At some point very soon we'll need to agree on the direction and either copy this code out to a shared extension or start to own and refactor it, but that's going to take a few weeks to iron out.

I agree, in terms of deliverables, let's get this merged in so that QE can test.

Then work backwards in improving it / getting everything refactored.

LGTM as long as we open up an issue to look back at this / possibly have some of this copypasta code in a central place.

cdrage (approved):

cdrage:
We should also test the `ssh2` requirement (test by building the image and installing it via OCI image)

PR #1600, opened 2025-05-07
chore: manually add macadam.js to the Containerfile

cdrage:
chore: manually add macadam.js to the Containerfile

### What does this PR do?

* Modifies the Containerfile so that the macadam.js library and binaries so that they can be used via OCI image install.

### Screenshot / video of UI

N/A, no UI changes.

Closes https://github.com/podman-desktop/extension-bootc/issues/1598

### How to test this PR?

1. Build the image (podman build) and push it to a container registry
2. Install via the container registry
3. See everything else works fine

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2024 Red Hat, Inc.
+# Copyright (C) 2024-2025 Red Hat, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -23,6 +23,11 @@ COPY LICENSE /extension/
 COPY packages/backend/icon.png /extension/
 COPY packages/backend/bootable.woff2 /extension/
 COPY README.md /extension/
+
+# We require the macadam.js binaries and library, so we will manually copy this over to the container image.
+# we rely on `pnpm build` before creating the container image, so we can safely assume that the macadam.js binaries are already present in the node_modules directory
+# and can copy them over to the container image.
+COPY node_modules/@crc-org/macadam.js /extension/node_modules/@crc-org/macadam.js
 
 FROM scratch
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 node_modules
+!node_modules/@crc-org/macadam.js


Containerfile
```diff
@@ -24,6 +24,11 @@ COPY packages/backend/icon.png /extension/
 COPY packages/backend/bootable.woff2 /extension/
 COPY README.md /extension/
 
+# We require the macadam.js binaries and library, so we will manually copy this over to the container image.
+# we rely on `pnpm build` before creating the container image, so we can safely assume that the macadam.js binaries are already present in the node_modules directory
+# and can copy them over to the container image.
+COPY node_modules/@crc-org/macadam.js /extension/node_modules/@crc-org/macadam.js
```
benoitf, line 30: hello, I'm not sure to follow why it's required ?

vite should include the dependency in the generated js file

feloy: But the binaries won't be included AFAIK

cdrage: That's right, and it won't execute correctly as the macadam js library relies on the module being there. 

So it's required to include it (similar to how it's done with ssh2 in the rhel vm extension).

benoitf: I still believe the way to go is relying on vite, not having some custom node_modules folder

see for example vite configuration of extension-ai-lab :
https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/backend/vite.config.js#L60-L79

benoitf: > That's right, and it won't execute correctly as the macadam js library relies on the module being there.
> 
> So it's required to include it (similar to how it's done with ssh2 in the rhel vm extension).

vite is the bundler, so we should rely on the bundler, not doing a manual copy in some Containerfile

the only thing to include should be the stuff generated in the `dist/` folder

deboer-tim: FWIW this exact code was approved and merged in the Macadam (now RHEL) extension, and due to dependency/timing issues we have to temporarily copy/paste it here instead of having a shared dependency or some other solution.

So: I totally agree we should investigate, but we need to be careful about having two different temporary solutions (harder to remove later or maintain in the meantime) or having different levels of review (b/c it functionally works, but macadam code is fresh and not meeting all our standards).

benoitf: @deboer-tim do you mean this line ? 
https://github.com/redhat-developer/podman-desktop-rhel-ext/blob/main/build/Containerfile#L31

cdrage: that's right @benoitf and with the new PR going in https://github.com/podman-desktop/extension-bootc/pull/1599 

we'll have to make sure that you also add ssh2 to this list as well @deboer-tim to be included in the image (if we get the error at all).

cdrage: FYI @deboer-tim and @benoitf I tried doing it the bundling route with vite this morning, and had issues when it came down to trying to access / use the included binary.. so I'm leaning towards the same solution that rhel VM extension had with the Containerfile. 

We *could* go the same route as Containerfile (download pnpm, do it via install, etc. all in the container), however we build ours a bit different (`pnpm build` BEFORE `podman build`), so either way it still installs the same module. Either way works for me.

deboer-tim (approved):
We need to follow up once the dust settles, but I'm approving so that we can get a solution in and move with other PRs/testing.

PR #1604, opened 2025-05-08
feat: enables virtual machine creation for linux

cdrage:
feat: enables virtual machine creation for linux

### What does this PR do?

* Modifies the if statement so that it shows for both mac and linux and
  not Windows
* Updates the macadam.js package to the most up-to-date / latest one.

### Screenshot / video of UI

Screenshot from 2025-05-08 13-45-37

Closes https://github.com/podman-desktop/extension-bootc/issues/1587

### How to test this PR?

Must test this out in the OCI image mode / don't use `pnpm watch`,
although `pnpm watch` will work, it's better to test this out in a
"production" environment.

1. Follow the README to install the required binaries
2. `pnpm build`
3. BUILD the image with `podman build`
4. Push the image somewhere
5. Install the OCI image under extensions.
6. Test out creating / starting / access the terminal on the VM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,26 @@ podman machine ssh
 rpm-ostree override reset --all
 ```
 
+**(Linux only) Unable to create virtual machine:**
+
+When creating a virtual machine on Linux, you may encounter a "macadam binary is missing" or "gvproxy binary is missing" message. This is because the binaries required have to be installed manually.
+
+Follow the below steps for the solution:
+
+**macadam binary:**
+
+1. Download the [macadam binary](https://github.com/crc-org/macadam/releases)
+2. `chmod +x macadam-linux-amd64`
+3. `sudo mkdir -p /opt/macadam/bin`
+4. `sudo mv macadam-linux-amd64 /opt/macadam/bin/macadam`
+
+**gvproxy binary:**
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -271,6 +271,26 @@ podman machine ssh
 rpm-ostree override reset --all
 ```
 
+**(Linux only) Unable to create virtual machine:**
+
+When creating a virtual machine on Linux, you may encounter a "macadam binary is missing" or "gvproxy binary is missing" message. This is because the binaries required have to be installed manually.
+
+Follow the below steps for the solution:
```
deboer-tim, line 278: Today someone said something was an 'egg chicken' problem, which I've always known as a 'chicken and egg' problem. I'm not sure who is correct 😆 but apparently we have the same: you always naturally write 'below steps' and I would have gone with 'steps below'. I guess this is fine. ;-)

deboer-tim (approved):
Code LGTM, thanks for getting this in!

README.md
benoitf: In French it's the egg first :-) https://fr.wikipedia.org/wiki/Paradoxe_de_l%27%C5%93uf_et_de_la_poule
In English the opposite https://en.wikipedia.org/wiki/Chicken_or_the_egg

PR #1607, opened 2025-05-09
chore: switch to npmjs package of crc-org binary

benoitf:
### What does this PR do?
It doesn't require anymore auth

>***note*** ensure to remove any custom .npmrc / specific auth for @crc-org package that you may still have locally

https://github.com/podman-desktop/extension-bootc/issues/1596

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -125,7 +125,7 @@
     "vitest": "^3.1.2"
   },
   "dependencies": {
-    "@crc-org/macadam.js": "0.0.1-202504301306-5ade487",
+    "@crc-org/macadam.js": "0.0.1-202505091153-d597ec3",
     "semver": "^7.7.1"
   }
 }
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,6 @@
 # Set update schedule for GitHub Actions
 
 version: 2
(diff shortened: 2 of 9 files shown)


packages/backend/package.json
```diff
@@ -125,7 +125,7 @@
     "vitest": "^3.1.2"
   },
   "dependencies": {
-    "@crc-org/macadam.js": "0.0.1-202504301306-5ade487",
+    "@crc-org/macadam.js": "0.0.1-202505091153-d597ec3",
```
axel7083, line 128: is the update needed to be in sync with the workflow update?

benoitf: unfortunately yes as it's now published to npmjs and not github packages (so we don't have previous versions)

but for the approval, need to see if it's still working with this new version

axel7083 (approved):

packages/backend/package.json
deboer-tim: Right now it needs to be in sync with rhel extension version too 😕, but agreed, one step at a time.

deboer-tim (approved):

packages/backend/package.json
benoitf: I made a PR in rhel extension as well

PR #1630, opened 2025-05-12
fix: wait for onmount and use fixed timer advance

### LINKED ISSUE #1624: Test failure in DiskImageDetailsBuild.spec.ts
### Bug description

All tests suddenly started failing with the following error:

```
 FAIL  src/lib/disk-image/DiskImageDetailsBuild.spec.ts > Refreshes logs correctly
AssertionError: expected "spy" to be called 2 times, but got 4 times
 ❯ vi.waitFor.timeout src/lib/disk-image/DiskImageDetailsBuild.spec.ts:98:46
     96|   await vi.waitFor(
     97|     () => {
     98|       expect(bootcClient.loadLogsFromFolder).toHaveBeenCalledTimes(2);
       |                                              ^
     99|     }
```
The issue is that the 2s refresh timer on the page is triggering 3 times before the waitFor (called every 250ms) is triggered.

### Operating system

macOS

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

The previous change switched to a fixed timer which is much more predictable and worked on main, but still fails on most builds. The failure is at least repeatable now, and shows the next problem: onMount is async, so there's no guarantee that the initial fetch or call to setInterval() has happened when we advance the timer and do the test.

This spys on setInterval to confirm that onMount() has executed and the initial fetch was done, then advances the timer once to make sure we've done exactly one step. Being called twice appears to be vitest artifact; does not happen when testing.

### Screenshot / video of UI

N/A

Fixes #1624.

### How to test this PR?

PR checks pass.

### DIFF
--- a/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.spec.ts
+++ b/packages/frontend/src/lib/disk-image/DiskImageDetailsBuild.spec.ts
@@ -94,10 +94,17 @@ test('Handles empty logs correctly', async () => {
 test('Refreshes logs correctly', async () => {
   vi.mocked(bootcClient.loadLogsFromFolder).mockResolvedValue(mockLogs);
   vi.mocked(bootcClient.getConfigurationValue).mockResolvedValue(14);
+  vi.spyOn(global, 'setInterval');
 
   render(DiskImageDetailsBuild, { folder: '/empty/logs' });
 
+  // make sure the timer is created (onMount has run)
+  await waitFor(() => {
+    expect(setInterval).toHaveBeenCalled();
+  });
+
   // verify we start refreshing logs
-  await vi.advanceTimersByTimeAsync(2500);
+  expect(bootcClient.loadLogsFromFolder).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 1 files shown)


jeffmaury (commented):
Code looks hard to read to me as setInterval is not mocked.

I would rather have:

const setIntervalSpy = vi.spy(global, 'setInterval');

and then expect(setIntervalSpy).to....

feloy (approved):
I agree with Jeff's comment that it is not easy to see that `setInterval` is spied on. 
LGTM otherwise

deboer-tim:
Updated to use `const setIntervalSpy ...` instead. 👍🏼

benoitf (approved):
there is probably a missing restoreMock or global restoreAllMocks as it'll be mocked for all the tests

deboer-tim:
> there is probably a missing restoreMock or global restoreAllMocks as it'll be mocked for all the tests

Will follow up with separate PR

PR #1651, opened 2025-05-15
fix: set macadam provider to the correct value

### LINKED ISSUE #1645: Error encountered when trying to create a VM on macos
### Bug description

Creating a VM from bootc image is currently impossible in the 1.9.0 release candidate on macOS, the following error is encountered when trying to create the VM:

Image

### Operating system

macOS

### Version

next (development version)

### Steps to reproduce

1. Pull an image from the Bootc examples page.
2. Build a raw/qcow2 image with it.
3. Try to start a VM with the previously build image from the bootc images page on macos
4. Notice the error encountered when trying to start up the VM.

### Relevant log output

```shell

```

### Additional context

_No response_

deboer-tim:
### What does this PR do?

See https://github.com/crc-org/macadam/issues/180 for background. Macadam is picking up the provider from .config/containers/containers.conf and fails on Mac when it uses libkrun.

This uses the same logic as the RHEL extension to explicitly set the provider option so that it doesn't get picked up from the file: https://github.com/redhat-developer/podman-desktop-rhel-ext/blob/main/src/extension.ts#L470

Does the correct thing for windows as well even though we don't support it yet. Added tests for all three platforms.

### Screenshot / video of UI

N/A

Fixes #1645.

### How to test this PR?

Set the provider in .config/containers/containers.conf to "libkrun" and have VM creation fail on 1.9, but same works fine after this fix.

### DIFF
--- a/packages/backend/src/macadam.ts
+++ b/packages/backend/src/macadam.ts
@@ -19,6 +19,7 @@ import * as macadam from '@crc-org/macadam.js';
 import { macadamName } from './constants';
 import { telemetryLogger } from './extension';
 import * as extensionApi from '@podman-desktop/api';
+import { isWSLEnabled } from './macadam/win/utils';
 
 interface StderrError extends Error {
   stderr?: string;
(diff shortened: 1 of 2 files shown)


packages/backend/src/macadam.ts
```diff
@@ -49,6 +50,16 @@ export class MacadamHandler {
           }
           options.imagePath = options.imagePath.replace(/^~\//, `${process.env.HOME}/`);
 
+          // force the correct provider so we're not dependent on .config/containers/containers.conf
+          let provider: 'wsl' | 'hyperv' | 'applehv' | undefined;
+          if (extensionApi.env.isWindows) {
+            provider = (await isWSLEnabled()) ? 'wsl' : 'hyperv';
```
feloy, line 56: For the `hyperv` I think this is a bug on RHEL ext we need to fix, as macadam does not support it yet AFAIK (https://github.com/crc-org/macadam/issues/60)

deboer-tim: Ok. We don't support Windows at all yet so I was just trying to keep the code in sync with RHEL ext, if there's an update there we'll pick it up later, or we start sharing this code somewhere in the future.

cdrage (approved):
Testing this on macOS, it propagates correctly.

Obviously unable to test via Windows, but since this inline with macadam / rhel vm extension, once bib supports .tar.gz, etc it'll work.

LGTM!

deboer-tim:
@Mergifyio backport 1.9.x

benoitf:
@Mergifyio backport 1.9.x

PR #1685, opened 2025-05-28
chore: add disclaimer that to test VM, have to add ssh key

cdrage:
chore: add disclaimer that to test, have to add ssh key

### What does this PR do?

* Adds a small disclaimer that only appears on mac or linux that you
  should add your SSH key + do output as raw or qcow2
* For this PR, I also did an alternative way of doing tests since I did
  not want to copy / paste and only be changing one thing. So in this PR
  the tests are "table-based".

### Screenshot / video of UI

Screenshot 2025-05-28 at 11 41 35 AM

Closes https://github.com/podman-desktop/extension-bootc/issues/1569

### How to test this PR?

1. pnpm watch
2. See that the disclaimer is shwon on the build page.

### DIFF
--- a/packages/frontend/src/Build.spec.ts
+++ b/packages/frontend/src/Build.spec.ts
@@ -943,3 +943,37 @@ test('purposely fail prereqs, expect an error to show going to build', async () 
     expect(errorMessage).toBeDefined();
   });
 });
+
+// The below mocks are the same across all tests.
+const setupPlatformMocks = (platform: string): void => {
+  vi.mocked(bootcClient.isMac).mockResolvedValue(platform === 'macOS');
+  vi.mocked(bootcClient.isWindows).mockResolvedValue(platform === 'windows');
+  vi.mocked(bootcClient.isLinux).mockResolvedValue(platform === 'linux');
+  vi.mocked(bootcClient.listHistoryInfo).mockResolvedValue(mockHistoryInfo);
+  vi.mocked(bootcClient.listBootcImages).mockResolvedValue(mockBootcImages);
+  vi.mocked(bootcClient.checkPrereqs).mockResolvedValue(undefined);
+  vi.mocked(bootcClient.buildExists).mockResolvedValue(false);
+  vi.mocked(bootcClient.inspectImage).mockResolvedValue(mockImageInspect);
+};
+
+// Test for macOS, Windows and Linux.
+test.each([
+  { platform: 'macOS', shouldShow: true },
+  { platform: 'windows', shouldShow: false },
+  { platform: 'linux', shouldShow: true },
+])('vm-disclaimer visibility on %s', async ({ platform, shouldShow }) => {
+  setupPlatformMocks(platform);
+
+  render(Build);
+
+  await vi.waitFor(() => {
+    expect(screen.getByLabelText('image-select')?.children.length).toBe(2);
+  });
+
+  const disclaimer = screen.queryByTestId('vm-disclaimer');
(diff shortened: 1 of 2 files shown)


packages/frontend/src/Build.svelte
```diff
@@ -1086,9 +1088,18 @@ $: if (availableArchitectures) {
...
+          {#if isMac || isLinux}
+            <p class="text-sm text-[var(--pd-content-text)] m-0 pt-0" data-testid="vm-disclaimer">
+              To test the virtual machine, add your public SSH key under <b>Users</b> in the interactive build config. Set the output format to RAW or QCOW2.
+            </p>
+          {/if}
+
           <!-- If on Linux, warn that during the build, credentials will be asked in order to run an escalated privileged build prompt -->
           {#if isLinux}
-            <p class="text-sm text-[var(--pd-content-text)] pt-1">
+            <p class="text-sm text-[var(--pd-content-text)] m-0 pt-0">
```
cdrage, line 1102: FYI, I changed the margin as well as the padding so it looks nicer. The default (from text-sm) was adding too much spacing.

packages/frontend/src/Build.spec.ts
```diff
@@ -943,3 +943,38 @@ test('purposely fail prereqs, expect an error to show going to build', async ()
     expect(errorMessage).toBeDefined();
   });
 });
+
+
+// The below mocks are the same across all tests.
+const setupPlatformMocks = (isMac: boolean, isWindows: boolean, isLinux: boolean) => {
```
deboer-tim, line 949: Suggestion: `const setupPlatformMocks = (platform: string) => {`
and
`vi.mocked(bootcClient.isMac).mockResolvedValue(platform == 'macOS'); ...`
and
`{ platform: 'macOS', shouldShow: true }, ...`
seems simpler?

deboer-tim (commented):
PR check (missing type) needs to be fixed.

I'd personally prefer the text was above (before) the button, but precedent already set with the existing Linux text. Nice update.

packages/frontend/src/Build.spec.ts
```diff
@@ -943,3 +943,39 @@ test('purposely fail prereqs, expect an error to show going to build', async ()
...
+  { platform: 'linux', shouldShow: true },
+])('vm-disclaimer visibility on %s', async ({ platform, shouldShow }) => {
+  setupPlatformMocks(platform);
+
+  render(Build);
+
+  await vi.waitFor(() => {
+    if (screen.getByLabelText('image-select')?.children.length !== 2) {
+      throw new Error('Image select options not loaded');
+    }
```
benoitf: usually waitFor is used with a expect(something).toBeSomething()

cdrage: updated! you're right, it works better and it's simpler to use expect rather than the if statement.

deboer-tim (approved):
Nit: since we have a note on all platforms, maybe
`<p class="text-sm text-[var(--pd-content-text)] m-0 pt-0>`
should be common with just a different message inside?

cdrage:
> Nit: since we have a note on all platforms, maybe `<p class="text-sm text-[var(--pd-content-text)] m-0 pt-0>` should be common with just a different message inside?

ACK, we'll have to refactor the technical debt in Build.svelte in the future with all the repeated classes :(

We'll have to put it in the custom tailwind config to use the classes (doing it statically via `const themed=`text-sm m-0` is not picked up when doing `pnpm build`).

Issue #1694, opened 2025-06-04
Clean up unused or not anymore needed dependencies and files in tests/playwright


odockal:
### Is your enhancement related to a problem? Please describe

We can remove electron dependency (electron is not needed in the extension anymore, see https://github.com/podman-desktop/podman-desktop/issues/11859) and also vitest + config which is not used as we are using playwright test runner. It should not affect unit tests, since vitest is defined in the root package.json file.

### Describe the solution you'd like

Remove electron dependency and also vitest + config. test build+test on CI.

PR #1695, opened 2025-06-04
chore(deps): remove unused dependencies and post install builds

odockal:
### What does this PR do?
Removes unused dependencies (vitest and electron) from `tests/playwright`, removes config for vitest and removes post install build script execution for any package. 
#1694


package.json
```diff
@@ -79,8 +79,7 @@
   "pnpm": {
     "overrides": {
       "vite>esbuild": "^0.25.0"
-    },
-    "neverBuiltDependencies": []
+    }
```
benoitf: I think it should be kept or add ignoredBuiltDependencies else we'll get a warning on each `pnpm install

```
╭ Warning ──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                               │
│   Ignored build scripts: @tailwindcss/oxide, cpu-features, esbuild, ssh2, svelte-preprocess, unrs-resolver.   │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.                      │
│                                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

odockal: True! I will add these the same way they are listed in ai-lab. As part of ignoredBuiltDependencies.

PR #1824, opened 2025-07-22
chore: update repository name

eqqe:
### What does this PR do?

Update repository name in Contributing instructions to help collaboration.


deboer-tim:
We should probably be updating these for all extensions to use the new local extensions support soon.

PR #1830, opened 2025-07-22
chore: add warning when cross-arch build on macOS

cdrage:
chore: add warning when cross-arch build on macOS

### What does this PR do?

* Adds a warning when building amd64 (we only support mac silicon)
* Tests added to compare on macOS vs Linux and Windows.

### Screenshot / video of UI

Screenshot 2025-07-22 at 4 42 00 PM

Closes https://github.com/podman-desktop/extension-bootc/issues/1735

### How to test this PR?

1. Be on macOS silicon
2. Try to build a disk image with amd64
3. See the warning under architecture selection before building.

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -27,6 +27,8 @@ import { goToDiskImages } from './lib/navigation';
 export let imageName: string | undefined = undefined;
 export let imageTag: string | undefined = undefined;
 
+const bootcExtensionGithubUrl = 'https://github.com/podman-desktop/extension-bootc';
+
 // Image variables
 let selectedImage: string | undefined;
 let existingBuild: boolean = false;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/Build.svelte
```diff
@@ -27,6 +27,8 @@ import { goToDiskImages } from './lib/navigation';
 export let imageName: string | undefined = undefined;
 export let imageTag: string | undefined = undefined;
 
+const bootcExtensionGithubUrl = 'https://github.com/podman-desktop/extension-bootc';
```
simonrey1, line 30: This could use the "repository" info from package.json (does not exist yet but could be added).

simonrey1:
I am having trouble to test your branch because I cannot select the x86 in the build page?

Screenshot 2025-07-23 at 09 31 49

jeffmaury (approved):
LGTM

cdrage:
> I am having trouble to test your branch because I cannot select the x86 in the build page?
> 
> Screenshot 2025-07-23 at 09 31 49

You will have to pull an amd64 image :) On the examples page on the top right of each example, you can pull the AMD64 version of it to test it out.

simonrey1:
> > I am having trouble to test your branch because I cannot select the x86 in the build page?
> > Screenshot 2025-07-23 at 09 31 49
> 
> You will have to pull an amd64 image :) On the examples page on the top right of each example, you can pull the AMD64 version of it to test it out.

Ok it works, but the warning was rather discrete and could use an error color?

packages/frontend/src/Build.svelte
benoitf: yes we should avoid defining such constants in the frontend

cdrage:
Screenshot 2025-07-25 at 3 33 50 PM

@eqqe Thanks for the suggestion, I did not add the warning colour at all! Makes sense to use the warning colour for a warning haha.

I'm going to set this to merge as I've now updated the colour + have 2 approvals + gotten feedback.

packages/frontend/src/Build.svelte
simonrey1: https://github.com/podman-desktop/extension-bootc/pull/1835

PR #2000, opened 2025-10-14
chore: use optional chaining

deboer-tim:
### What does this PR do?

Changes two instances to use optional chain expressions. More concise, and should unblock #1993.

### Screenshot / video of UI

N/A

See new aslant error in #1993.

### How to test this PR?

Tests pass, confirm logic is unchanged.

### DIFF
--- a/packages/backend/src/container-utils.ts
+++ b/packages/backend/src/container-utils.ts
@@ -99,7 +99,7 @@ export async function deleteOldImages(engineId: string, currentImage: string): P
   try {
     // List all the images and check to see if it exists
     const images = await extensionApi.containerEngine.listImages();
-    if (!images || images.length === 0) {
+    if (!images.length) {
       return;
     }
 
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -182,7 +182,7 @@ async function validate(): Promise<void> {
     return;
   }
 
-  if (!buildType || buildType.length === 0) {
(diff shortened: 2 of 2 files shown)


packages/backend/src/container-utils.ts
```diff
@@ -99,7 +99,7 @@ export async function deleteOldImages(engineId: string, currentImage: string): P
   try {
     // List all the images and check to see if it exists
     const images = await extensionApi.containerEngine.listImages();
-    if (!images || images.length === 0) {
+    if (!images.length) {
```
benoitf: it does not look equivalent to the original solution

equivalent is `if (!images?.length) {`

deboer-tim: Sorry, had that before and not sure how I lost the char before commit. Fixed now.

simonrey1 (approved):
LGTM

question: did you use `pnpm lint:fix`?

deboer-tim:
> question: did you use `pnpm lint:fix`?

No, I just made the change myself and (out of curiosity) tried Cursor for one and it was making the same change.

Turns out it's good I didn't use `lint:fix`. I just checked and it changes:
  `if (!images || images.length === 0) {`
to
  `if (images?.length === 0) {`
which changes the logic by missing the ! case. :-/

benoitf:
@deboer-tim I think you need to amend the PR to trigger again the semantic PR check that was probably offline

PR #2017, opened 2025-10-20
chore: fix lock file

deboer-tim:
### What does this PR do?

pnpm-lock.yaml became broken after multiple dependabot updates: `The lockfile at ".../pnpm-lock.yaml" is broken: duplicated mapping key (3415:3)`

Deleted and used `pnpm install` to recreate. Should unblock #2010 and #2011.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

Regular PR checks.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3412,15 +3412,6 @@ packages:
       svelte:
         optional: true
 
-  svelte-eslint-parser@1.4.0:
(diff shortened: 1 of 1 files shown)


deboer-tim:
> too many changes, it should only contain the duplicated lines being removed

Fixed, only contains the duplicate removal now.

benoitf (approved):
thanks

Issue #2051, opened 2025-11-06
Use Github issue types


deboer-tim:
### Is your enhancement related to a problem? Please describe

Now that they exist, we should use GitHub issue types instead of the existing labels.

### Describe the solution you'd like

As discussed on team meeting, we should:
- [x] update the issue templates to use Github's issue types instead of labels: #2052 #2056 #2077
- [x] update existing issues from labels to types
- [x] merge enhancements into feature requests, since they are basically identical and we'll only keep one type: #2052
- [x] eventually, delete the labels #2078

PR #2056, opened 2025-11-06
chore: fix epic issue type and label

deboer-tim:
### What does this PR do?

Adding `type: epic` to the issue template not only didn't work, the kind/epic label doesn't show up anymore either. I don't have access to the org settings but the epic type appears to include the emoji - assuming it was copied from the kind and this should fix it.

### Screenshot / video of UI

N/A

Fix for part of #2051.

### How to test this PR?

Merge, confirm type and label work.


MarsKubeX (approved):

jeffmaury (approved):
Maybe we should reserve emojis to labels

PR #2075, opened 2025-11-14
chore(docs): update the correct url for the dashboard

cdrage:
chore(docs): update the correct url for the dashboard

On clicking, this should actually take you to the readme URL rather than
the 404 page.

N/A, found during event.

### How to test this PR?

N/A, see markdown.

### DIFF
--- a/packages/frontend/src/lib/FirstImage.svelte
+++ b/packages/frontend/src/lib/FirstImage.svelte
@@ -11,6 +11,7 @@ let pullInProgress = $state(false);
 let displayDisclaimer = $state(false);
 
 const exampleImage = 'registry.gitlab.com/fedora/bootc/examples/httpd:latest';
+const exampleImageReadmeUrl = 'https://gitlab.com/fedora/bootc/examples/-/tree/main/httpd';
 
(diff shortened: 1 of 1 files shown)


feloy (approved):

packages/frontend/src/lib/FirstImage.svelte
```diff
@@ -11,6 +11,7 @@ let pullInProgress = $state(false);
 let displayDisclaimer = $state(false);
 
 const exampleImage = 'registry.gitlab.com/fedora/bootc/examples/httpd:latest';
+const exampleImageReadmeUrl = 'https://gitlab.com/fedora/bootc/examples/-/tree/main/httpd';
```
benoitf, line 14: nit: I would expect EXAMPLE_IMAGE_README_URL for immutable basic type constants. (but then it's not consistent with the previous constant)

cdrage: I agree :( we need to do this change through-out all of the bootc code, the standards we have for constant's aren't up to date (much easier to read capitalized).

LGTM to merge this (for now), we'll make the change for const

Issue #2101, opened 2025-11-25
provide build using Containerfiles


benoitf:
### Is your feature related to a problem? Please describe

Today, the build is done purely inside GitHub runners (having the Node.js dependency, package manager, etc)

Some other projects are providing Containerfile (one for the builder and one for the extension) allowing to quickly build the extension without any dependency installed on the system (except the OCI runtime like podman, docker, etc)

example:
https://github.com/podman-desktop/extension-kubernetes-dashboard/tree/main/build
https://github.com/podman-desktop/extension-apple-container/tree/main/build

(and see usage inside .github/workflows)

### Describe the solution you'd like

Introduce a build directory with Containerfile to have a build being possible without requiring GitHub runner / workflow so we really know what is provided to the building tools steps.

### Describe alternatives you've considered

keep it like it is

deboer-tim:
Should use ubi-10.

simonrey1:
Unassigning myself in favor of Tim as he already started to work on this issue.

cdrage:
PR(s) have been merged in. Closing this issue as 1.11 is now out and built using the new Containerfiles.

PR #2118, opened 2025-12-03
chore: switch to container build

deboer-tim:
### What does this PR do?

The final output of the build was an OCI container, but it was purely used for packaging. This moves the Containerfile to the /build folder and does the full pnpm build using a builder image.

Images published are extension-bootc-builder and extension-bootc, switching from the previous podman-desktop-extension-bootc to match what newer extensions have done and make the switch obvious.

First part: leaves existing container file in root for e2e tests to run, once workflow has successfully run once we can reuse the latest builder image instead, and remove the Containerfile at root.

### Screenshot / video of UI

N/A

First part of #2101.

### How to test this PR?

Merge, make sure builds complete and compare images.

### DIFF
--- a/.github/workflows/build-next.yaml
+++ b/.github/workflows/build-next.yaml
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2023-2024 Red Hat, Inc.
+# Copyright (C) 2023-2025 Red Hat, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


build/Containerfile.builder
simonrey1, line 1: I tried to build this Containerfile.builder but I am seing the following error:

Image


Image

I tried to do a  `git clean -xdf`

.github/workflows/build-next.yaml
```diff
@@ -23,37 +23,111 @@ on:
...
+          fetch-depth: 2 # Ensure we have at least one previous commit for diff check
+
+      - name: Check for builder-related changes
+        id: check
+        run: |
+          if git diff --name-only HEAD^ HEAD | grep -E '^(package.json|pnpm-lock.yaml|build/Containerfile.builder|.github/workflows/build-next.yaml)$'; then
+            echo "builder_required=true" >> $GITHUB_OUTPUT
+          else
+            echo "builder_required=false" >> $GITHUB_OUTPUT
+          fi
```
axel7083, line 43: I'm a bit septic, are we sure we will not in a near future have this not covering all cases?

.github/workflows/build-next.yaml
```diff
@@ -23,37 +23,111 @@ on:
...
+          registry: ghcr.io/${{ github.repository_owner }}
 
-      - name: Run Build
-        run: pnpm build
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@v3
+        with:
+          subject-name: ghcr.io/${{ github.repository_owner }}/extension-bootc-builder
+          subject-digest: ${{ steps.push-to-ghcr.outputs.digest }}
+          push-to-registry: true
```
axel7083, line 88: Very interesting 👁️

.github/workflows/build-next.yaml
```diff
@@ -23,37 +23,111 @@ on:
...
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: build extension image
+        id: extension-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: extension-bootc
+          tags: next ${{ github.sha }}
```
axel7083, line 106: a tag with space?

build/Containerfile.builder
```diff
@@ -0,0 +1,28 @@
+#
```
axel7083, line 1: out of the context of this PR, but we need to revise the usage of a builder container, multi stage container exists, and we could use caching mechanism to make them as fast, and avoid publishing a builder image

.github/workflows/build-next.yaml
feloy: this is a list of tags:
- next
- the sha of commit

benoitf (changes requested):
build doesn't work

in builder, adding the line 
```
 # copy the application files to the /opt/app-root/extension-source directory
 WORKDIR /opt/app-root/extension-source
+RUN mkdir -p /opt/app-root/extension-source
```

fixes the issue, but then the Containerfile is not working. Is it up-to-date ?

for example it tries to copy files that do not exist (as it's inside some subfolders) (like *.png, scripts, etc)

.github/workflows/build-next.yaml
deboer-tim: Agreed, but following the pattern from Apple and RHEL extensions. So I'd prefer to defer due to 'existing pattern' and I think we need a 'build epic' for future sprint where someone would look at multi-stage build, sharing workflow jobs, etc.

deboer-tim:
> build doesn't work

Sorry, this PR contained lots of crap, and I clearly hadn't run what I pushed locally. To top it off I applied suggestions from the automated PR review that actually broke it further.

I've replaced most of the file copying sections and learned from what other extensions did, so I've force-pushed to start over. This commit successfully builds both images for me, and I've confirmed that the final contents match the last bootc release image.

build/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+      cp -r packages/backend/media/ /opt/app-root/extension/media
+
+COPY LICENSE /opt/app-root/extension/
+COPY README.md /opt/app-root/extension/
+
+# TEMPORARY. Permanent fix will be in the future when we can add all of this to vite script.
+# We require the macadam.js binaries and library, so we will manually copy this over to the container image.
+# we rely on `pnpm build` before creating the container image, so we can safely assume that the macadam.js binaries are already present in the node_modules directory
+# and can copy them over to the container image.
+COPY node_modules/@crc-org/macadam.js /opt/app-root/extension/node_modules/@crc-org/macadam.js
```
jeffmaury, line 51: **question:** I am a little bit concerned why those dependencies are handled manually and not through pnpm like the RHEL extension ?

deboer-tim: Me too, but this is unchanged from the current containerfile. As someone we know would say 'one PR one change'. :)

I see what the RHEL extension did and can replicate in a following PR.

jeffmaury: Not sure to understand as Containerfile(s) are new in this PR ?

simonrey1: > Not sure to understand as Containerfile(s) are new in this PR ?

There is an existing Containerfile at the root of the repository

cdrage: non-blocker, but a concern. 

But any changes to the Containerfile that's outside-the-norm we should also check to see if it's correctly adding the arm64 madacam binary when building an arm64 container image.

Same for amd64, etc.

There were issues in the past where you would build this container image with `podman build --platform linux/amd64` and because you are building it on macOS.. it adds the arm64 binary, etc.

deboer-tim:
> maybe QE should validate (or validate the image that will be built on CI before cutting a tag)

Yes, I want to merge and validate the published image, then give QE a chance to try before we start a release.

build/Containerfile
deboer-tim: +1 to this. I think I have a changeset that replicates what the RHEL extension build does, but that path changes where node_modules ends up in the final image and the structure of it (plus adds a few Mb). It would change this PR from 'it generates the identical extension as today, but we should still test' to 'the build output is fundamentally different so the only way to validate is testing on every platform'. I've opened #2136 to track this as a separate issue.

cdrage:
Was able to build too and load the plugin! Tested on macOS. 

No concerns with regards to building (it's the same if not similar to what we currently have).

LGTM!

cdrage (approved):

PR #2116, opened 2025-12-02
chore: test cleanup

deboer-tim:
### What does this PR do?

Cleans up some issues in tests:
- Adds missing getUidGid and listVMs mocks.
- Simplify ResizeObserver mocking using the same pattern as Podman Desktop renderer.
- Import correct beforeEach.
- Import node types to avoid NodeJS import warnings.

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

Fixes #2115.

### How to test this PR?

Code looks ok; tests green.

PR #2124, opened 2025-12-04
chore(deps): bump vitest to v4

axel7083:
### What does this PR do?

Vitest 3 seems to have some security CVE in one of its package, let's address this by switching to Vites v4: this is not fun to do :) 

I splited in several commits all changes but one are for tests only, the only thing I had to change was to update the following classes constructors
- `BootcApiImpl`
- `MacadamHandler`

To get the `telemetryLogger` from the constructor instead of importing it from `extension.ts`. 

> ⚠️  If this is too much I can try to split this PR

### Screenshot / video of UI

N/A

Fixes https://github.com/podman-desktop/extension-bootc/issues/2107

### How to test this PR?

- CI/CD should be 🟢


packages/backend/__mocks__/@crc-org/macadam.js
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
```
axel7083: ⚠️ git is not detecting nicely, but this is not a rename 

I deleted `__mocks__/@podman-desktop/api.js` because it was useless

I created `packages/backend/__mocks__/@crc-org/macadam.js` to mock the `@crc-org/macadam.js` nicely because it was a pain in some places

packages/backend/src/api-impl.spec.ts
```diff
@@ -16,50 +16,62 @@
...
-});
-
-vi.mock('@crc-org/macadam.js', () => {
-  const mockInstance = {
-    init: vi.fn(),
-    createVm: vi.fn(),
-    listVms: vi.fn(),
-  };
-  return {
-    Macadam: vi.fn(() => mockInstance),
```
axel7083, line 52: This syntax is not valid in vitest v4

packages/backend/src/extension.spec.ts
```diff
@@ -86,23 +86,10 @@ vi.mock('@podman-desktop/api', async () => {
   };
 });
 
-vi.mock('../package.json', () => ({
```
axel7083, line 89: There is twice the same mock, removing it to avoid issues here at L89 and

https://github.com/podman-desktop/extension-bootc/blob/87b0342ba5627b294cf1b06ff3f7006bf2c65d16/packages/backend/src/extension.spec.ts#L35-L39

packages/frontend/src/stores/rpcReadable.ts
```diff
@@ -28,8 +28,8 @@ export function RPCReadable<T>(
   // For example, you can pass in a custom function such as "getPullingStatuses".
   updater: () => Promise<T>,
 ): Readable<T> {
-  let timeoutId: NodeJS.Timeout | undefined;
-  let timeoutThrottle: NodeJS.Timeout | undefined;
+  let timeoutId: number | undefined;
+  let timeoutThrottle: number | undefined;
```
axel7083: typecheck is failing: NodeJS cannot be accessed from the frontend

deboer-tim:
Merge #2116 first and rebase?

axel7083:
> Merge #2116 first and rebase?

Done ✅

deboer-tim (approved):
Code LGTM, tests pass. 👍🏼

packages/backend/vite.config.js
```diff
@@ -55,6 +55,17 @@ const config = {
     emptyOutDir: true,
     reportCompressedSize: false,
   },
+  test: {
+    include: ['**/*.{test,spec}.?(c|m)[jt]s?(x)', '../shared/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
```
jeffmaury, line 59: **question:** why is shared included here ?

axel7083: I moved the `vitest.config.js` in the `vite.config.js`, seems to have some problem when they are not in the same place

jeffmaury: Aren't you mixing backend and shared together with this ?

axel7083: The shared is already mixed with the backend, as of today here is the `packages/backend/vitest.config.js#L26`

https://github.com/podman-desktop/extension-bootc/blob/51d0426df3aaa14b72509750b1c6baf4aa996057/packages/backend/vitest.config.js#L26

It is technically not nice as the shared package has its own vitest.config.js 

https://github.com/podman-desktop/extension-bootc/blob/51d0426df3aaa14b72509750b1c6baf4aa996057/packages/shared/vitest.config.js#L26

Feel to me that all of this are copy paste from other repo, and would probably need more cleanup, but as of this PR, I don't change the include logic

PR #2179, opened 2025-12-19
feat(macadam): only install macadam when vm features are used

### LINKED ISSUE #2175: BootC extension throws errors in the console: Error: component not initialized on Mac OS
### Bug description

If extension is added to the distribution and start up, bootc extension is throwing errors in the console log (maybe macOS only).

`[bootc] Error: component not initialized. You must call init() before`

### Operating system

macOS

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

cdrage:
feat(macadam): only install macadam when vm features are used

### What does this PR do?

Macadam is used to run VM's, this was previously activated on boot and would
install macadam immediately by showing a sudo prompt. This PR adds
"lazy" installing functionality where we only prompt if VM features are
clicked / used.

This PR:
* Moves the init functionality to only happen when creating a VM
* The VM monitoring loop is triggered either at the beginning (if
  macadam is already installed) or after create VM is used which would
  be the first point a VM is accessed / used
* Adds documentation as well as screenshot

### Screenshot / video of UI

**Video shows functionality of it installed and not installed:**

https://github.com/user-attachments/assets/75ed6402-b380-47ce-baae-73e349cb5fac

Closes https://github.com/podman-desktop/extension-bootc/issues/2171
Closes #2175

### How to test this PR?

1. REMOVE MACADAM (macOS):

```sh
~ $ whereis macadam
macadam: /opt/macadam/bin/macadam
~ $ sudo rm -rf /opt/macadam
Password:
```

If you are on Linux, it must be removed at: `/usr/local/bin/macadam`

2. Start bootc extension, you should NO LONGER receive the sudo prompt.

3. Go to **Disk Images > Create VM**

4. Expect the sudo prompt to appear to install

5. Be able to create virtual machine / list VM functionality like usual.


packages/backend/src/extension.ts
```diff
@@ -178,17 +187,29 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
+    const macadamBinaryPath = extensionApi.env.isMac ? MACADAM_BINARY_PATH_MACOS : MACADAM_BINARY_PATH_LINUX;
+    if (fs.existsSync(macadamBinaryPath)) {
+      try {
+        await macadam.init();
+        macadamInitialized = true;
+      } catch (error) {
+        console.error('Error initializing macadam', error);
+      }
+
+      monitorMachines(provider, extensionContext).catch((error: unknown) => {
```
feloy: I think it would better separate concerns to fire an event `initialized` here, and have an `initialized.on(() => monitorMachines)` at the place of the original `monitorMachines()`

cdrage:
Updated! Great suggestion @feloy I switched it to use events emitter instead to retrigger the monitoring.

packages/backend/src/extension.ts
```diff
@@ -178,17 +190,30 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
+        console.error('Error while monitoring machines', error);
+      });
     });
+
+    // Only initialize and start monitoring if macadam binary is already installed.
+    // This avoids prompting for sudo on extension activation (macOS).
+    // If not installed, init() will be called lazily when user performs a VM operation.
+    const macadamBinaryPath = extensionApi.env.isMac ? MACADAM_BINARY_PATH_MACOS : MACADAM_BINARY_PATH_LINUX;
+    if (fs.existsSync(macadamBinaryPath)) {
```
feloy: I'm a little concerned about having this logic here, because we will need to duplicate it in all extensions using the macadam.js library, and because the whole logic is splitted between the library and here.

I would propose to change the signature of the `init()` method of the library, to:

```
interface InitOptions {
  allowPrivilegedOperations: boolean;
}

init(options?: InitOptions): Promise<boolean>
```

The `init` would return true only if it was able to do its job, considering the `allowPrivilegedOperations` option: 
- if the binaries need to be installed and allowPrivilegedOperations is false => returns false
- otherwise, returns true

The extensions would call `init({ allowPrivilegedOperations: false })` at startup, then `init({allowPrivilegedOperations: true})` for other operations.

@cdrage @benoitf WDYT? Is changing the signature/behaviour of `init` a concern?

cdrage: IMO if we can have the logic with regards to installation in the macadam library, that would be way more ideal.

Honestly, my opinion is that we need the binary installed anyways in order to actually list all available machines with the monitoring logic.

However, if we're able to get this implemented in the library in time, that'd be great. We also have this issue too: https://github.com/crc-org/macadam.js/issues/50 were we cannot get the export / hardcoded value of the binary locations (I had to add this code logic in the bootc extension code).

feloy: I can see 3 different solutions:
- the library shares the paths of the binaries, and the extensions check if the binaries are available
- the library provides a function `areBinariesAvailable(): boolean`, to be used by extensions
- the library provides the `init` function as described above, and the extensions calls it with different options, depending on when it is called.

I think that the third one is more "reactive", in the sense that the extensions call this function by declaring what can be done, then the library makes what it can with these authorizations, and the extensions need to react on any possibilities.

Otherwise, between the first two options, the second one has my preference

cdrage: I'm on board then with `the library provides the init function as described above, and the extensions calls it with different options, depending on when it is called.` and I can update this PR with the following new library code.

It would be great to have the `areBinariesAvailable` available (esp if we are supporting flatpak users too..) / having https://github.com/crc-org/macadam.js/issues/50 fixed, then I could use this logic instead of the hardcoded linux values I have at the moment.

feloy: I've had a try on solution 3 with rhel ext, but that would change the extension's logic too much. Let's start with solution 2 for the moment: I'll create a PR to fix https://github.com/crc-org/macadam.js/issues/50

deboer-tim: At some point I think we really should do a shared extension that manages Macadam as a CLI tool (allows users to directly see it, update, or uninstall which they can't do today) and provides or exposes this basic api for other extensions to use it. For now I'm ok updating `init` though as it's far less work.

cdrage:
@feloy thank you for the changes! with the updated macadam library, I'm now using the `areBinariesAvailable` function.

packages/backend/src/extension.spec.ts
```diff
@@ -107,6 +121,65 @@ const fakeContext = {
   storagePath: os.tmpdir(),
 } as unknown as podmanDesktopApi.ExtensionContext;
 
+describe('test ensureMacadamInitialized doesnt double init', () => {
+  beforeEach(() => {
+    vi.spyOn(fs.promises, 'readFile').mockImplementation(() => {
```
jeffmaury: **praise:** mockResolvedValue ?

packages/backend/src/extension.spec.ts
```diff
@@ -194,3 +267,77 @@ test('check command triggers webview and redirects', async () => {
   expect(podmanDesktopApi.navigation.navigateToWebview).toHaveBeenCalled();
   expect(postMessageMock).toHaveBeenCalledWith({ body: 'build/latest', id: 'navigate-build' });
 });
+
+describe('lazy macadam initialization', () => {
+  beforeEach(() => {
+    vi.spyOn(fs.promises, 'readFile').mockImplementation(() => {
+      return Promise.resolve('<html></html>');
```
jeffmaury: **praise:** mockResolvedValue ?

packages/backend/src/extension.ts
```diff
@@ -234,6 +253,27 @@ export async function getConfigurationValue<T>(property: string): Promise<T | un
...
+    await macadam.init();
+    macadamInitialized = true;
+
+    // Emit initialized event to trigger monitoring
+    macadamEvents.emit(MACADAM_INITIALIZED_EVENT);
+  } catch (error) {
+    console.error('Error initializing macadam', error);
+    throw error;
+  }
+}
```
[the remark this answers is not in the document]
feloy: In my opinion this race condition cannot happen, as the UI does not permit this

feloy (approved):
LGTM codewise

PR #2217, opened 2026-01-13
chore: svelte 5 migration

### LINKED ISSUE #1295: Svelte 5 migration

We use Svelte v5 now, should slowly start to migrate all components from v4.

Migrate all code.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Svelte 5 migration died off a bit before it was complete. This just wraps up most of the remaining work enough to close the issue.

There are still two components that haven't been migrated:
- Router - hasn't been done in PD either, and we've talked about replacing it.
- Build page - will be a bunch of work, can be handled separately some time.

### Screenshot / video of UI

N/A

Fixes #1295.

### How to test this PR?

Existing tests should pass, no regressions to these parts of the UI.

### DIFF
--- a/packages/frontend/src/lib/BootcStatusIcon.svelte
+++ b/packages/frontend/src/lib/BootcStatusIcon.svelte
@@ -41,7 +41,7 @@ let IconComponent = $derived(icon);
     title={status}>
     {#if status === 'running' || status === 'creating' || status === 'deleting'}
       <Spinner size="1.4em" />
-    {:else}
+    {:else if IconComponent}
       <IconComponent {size} {solid} ></IconComponent>
     {/if}
   </div>


packages/frontend/src/lib/BootcStatusIcon.svelte
```diff
@@ -38,7 +42,7 @@ $: solid =
     {:else if typeof icon === 'string'}
       <span class={icon} aria-hidden="true"></span>
     {:else}
-      <svelte:component this={icon} size={size} solid={solid} />
+      <icon size={size} solid={solid} ></icon>
```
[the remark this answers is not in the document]
deboer-tim: We're using Svelte 5, you shouldn't use `<svelte:component>` anymore.

packages/frontend/src/lib/examples/ExampleDetails.svelte
gastoner, line 26: Thinking if we can also use derived with await in bootc? (not sure TBH, in Podman Desktop this rule is enabled so whole onMount could be removed then)

gastoner: But probably also an another issue

deboer-tim: Good catch. I tried it and it is an improvement, so I included it here, although not a huge win due to the other changes required.

Also found the icon wasn't working (and I hadn't actually been testing my dev extension) so fixed that as well.

packages/frontend/src/lib/BootcStatusIcon.svelte
```diff
@@ -5,18 +5,24 @@ import type { Component } from 'svelte';
...
 // any other status will result in a standard outlined box
-export let status = '';
-export let icon: Component | string | undefined = undefined;
-export let size = 20;
+interface Props {
+  status?: string;
+  icon?: Component;
+  size?: number;
+}
+let { status = '', icon, size = 20 }: Props = $props();
```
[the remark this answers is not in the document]
deboer-tim: icon is always set so not really an issue, but I added this check.

PR #2223, opened 2026-01-14
feat: link to image details

### LINKED ISSUE #2206: With new release: update minimum requirement for PD to 1.25.0 due to Node 24 and sudo-prompt
### Is your feature related to a problem? Please describe

In Podman Desktop core we updated to Node 24.

This extension we updated to Node 24 too.

We use sudo-prompt within the extension for using escalated privileges, which was also updated to Node 24 (https://github.com/podman-desktop/podman-desktop/issues/15548)

We will require the usage of Podman Desktop 1.25.0 in order to use the sudo-prompt escalation for Windows / macOS / Linux. 

_No response_

### Additional context

_No response_

### LINKED ISSUE #2218: Make the user understand he can build his own bootc image using `Images > Build an image`
### Is your feature related to a problem? Please describe

> Got feedback from a user

As explained in https://github.com/podman-desktop/extension-bootc?tab=readme-ov-file#usage we can go inside the `Images > Build an image` and select a containerfile.

Once built it will appear inside the `Bootc > Images`, but from the bootc webview, it is unclear how to add our own image.

We could have a button in `Bootc > Images` that redirect to the `Images > Build an image`, we could duplicate the `Build` that already exists in `Images` page

We also could have in the dashboard, a suggestion to use the `Images > Build an image` page to create bootc images

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Adds a build button to the Images page, that links to the standard Image build page.

Renamed the existing gotoImageBuild to gotoDiskImageBuild (first commit) since it was poorly named and not following the convention.

### Screenshot / video of UI

Screenshot 2026-01-14 at 1 53 48 PM

Fixes #2218.
Fixes #2206.

### How to test this PR?

Go to the Bootc > Images and make sure the Build button links to the Podman Desktop image build page.

### DIFF
--- a/packages/backend/src/api-impl.ts
+++ b/packages/backend/src/api-impl.ts
@@ -355,7 +355,13 @@ export class BootcApiImpl implements BootcApi {
   }
 
   async openImageBuild(): Promise<void> {
-    return await podmanDesktopApi.navigation.navigateToImageBuild();
+    try {
+      return await podmanDesktopApi.navigation.navigateToImageBuild();
+    } catch (err) {
+      // navigateToImageBuild was added in PD 1.25
+      console.error('Build page navigation failed, going to Images instead', err);
+      return await podmanDesktopApi.navigation.navigateToImages();
+    }
   }
 
   async generateUniqueBuildID(name: string): Promise<string> {


packages/backend/src/api-impl.ts
```diff
@@ -354,6 +354,10 @@ export class BootcApiImpl implements BootcApi {
     return await podmanDesktopApi.navigation.navigateToImage(id, engineId, tag);
   }
 
+  async openImageBuild(): Promise<void> {
+    return await podmanDesktopApi.navigation.navigateToImageBuild();
+  }
```
axel7083, line 365:
```suggestion
  async openImageBuild(): Promise<void> {
    if version >= 1.25.0 
        return await podmanDesktopApi.navigation.navigateToImageBuild();
    else 
        return await podmanDesktopApi.navigation.navigateToImages();
   
  }
```

We have a lot of users that may not have the latest version, we should handle the case where the api is not available :/

packages/backend/package.json
```diff
@@ -7,7 +7,7 @@
   "publisher": "redhat",
   "license": "Apache-2.0",
   "engines": {
-    "podman-desktop": ">=1.18.0"
+    "podman-desktop": ">=1.25.0"
```
axel7083, line 10: We should keep it broad, but have a fallback on the navigate, so users on older version can still benefits the new feature added (other than the navigate)

packages/backend/src/api-impl.ts
deboer-tim: I'd like it to be available on older Podman Desktop versions, but we were going to have to set 1.25 as the minimum anyway due to #2206, so I wasn't going to bother with this.

axel7083: I would still add the if else, because I don't think we prevent anyone from installing on older version

deboer-tim: apiVersion was also added in 1.25 :) so I just added a simple try/catch.

Issue #871, opened 2024-09-30
Details page should link to Images


deboer-tim:
### Is your enhancement related to a problem? Please describe

The Disk Image Summary tab shows the image as plain text.

### Describe the solution you'd like

As suggested in comments of #866, the image listed in the Summary tab should be a link to the Images page.

PR #2231, opened 2026-01-16
chore: add telemetry to image navigation

deboer-tim:
### What does this PR do?

Adds telemetry when navigating to an image. We do it for all the others, should do this one too.

### Screenshot / video of UI

N/A

Part of #871.

### How to test this PR?

Just PR checks.


axel7083 (approved):
I would prefer to see the telemetry stay in the backend, but as the functions are already exposed, I'm approving! LGTM

https://github.com/podman-desktop/extension-bootc/blob/cc5c66696968b2afeb9c0cffd3a1d615618b8013/packages/shared/src/BootcAPI.ts#L63-L64

PR #2246, opened 2026-01-20
chore: integrate ExamplesCard tests

### LINKED ISSUE #2245: Integrate ExamplesCard tests
### Is your feature related to a problem? Please describe

I noticed that we have ExamplesCard.ts, which contains tests for the ExamplesCard.svelte, but doesn't end with .spec.ts so the tests are never run.

Clean up any issues with the test and rename the file so that it is executed.

_No response_

### Additional context

_No response_

deboer-tim:
### What does this PR do?

Renames the file to end with .spec.ts so that the tests are correctly run in the build.

- First failed due to missing getArch(), fixed.
- Second test failed, removed b/c the identical test is already in ExampleCard.spec.ts (note singular), and that's where it should be.
- Third test failed b/c the card is empty if there are no examples in the category, fixed by adding a message to the .svelte (and improving the text).

### Screenshot / video of UI

N/A

Fixes #2245.

### How to test this PR?

Regular PR checks.

### DIFF
--- a/packages/frontend/src/lib/examples/ExamplesCard.svelte
+++ b/packages/frontend/src/lib/examples/ExamplesCard.svelte
@@ -17,5 +17,8 @@ let { category, examples }: Props = $props();
         <ExampleCard {example} />
       {/each}
     </div>
+    {#if examples.length == 0}
+      There are no examples in this category.
+    {/if}
   </div>
 </Card>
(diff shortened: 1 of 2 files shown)


packages/frontend/src/lib/examples/ExamplesCard.svelte
```diff
@@ -17,5 +17,8 @@ let { category, examples }: Props = $props();
         <ExampleCard {example} />
       {/each}
     </div>
+    {#if examples.length == 0}
+      There are no examples in this category.
+    {/if}
```
cdrage, line 22: 👍 Yay enabling this.

Thanks for the quick fix / showing if there's zero examples. Testing it locally I had no issues with it passing. (the examples did not show when I purposely removed the json entries too).

LGTM

PR #2358, opened 2026-03-10
chore(test): fixing race condition in e2e tests on windows

cbr7:
### What does this PR do?
Attempts to fix the e2e tests when encountering a race condition on windows OS.


cbr7:
Started a windows test run at https://github.com/podman-desktop/extension-bootc/actions/runs/22901747123

PR #2371, opened 2026-03-12
fix(build): cache checkPrereqs and debounce validation to fix windows timeouts

### LINKED ISSUE #1701: Error: Timeout encountered in bootc build page on windows
### Bug description

Currently when running the automated tests on windows cicd there is a timeout error encountered in the bootc build form which disables the `Build` button. It's not clear what throws this error or why it's only really seen in automatic tests but not manual and it's only encountered on Windows OS.

Image

### Operating system

window 11

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

cdrage:
fix(build): cache checkPrereqs and debounce validation to fix windows timeouts

### What does this PR do?

Cache `checkPrereqs()` results and debounce form validation to fix the
timeout error encountered on the bootc build page on Windows.

The cause was that `checkPrereqs()` was being called on every form
change (including each keystroke when typing the output folder path).

At one point.. it would call it 17 times while typing it in, so in a CI
scenario, it was validating WAY too much.

On Windows/WSL, each `checkPrereqs()` call runs `podman machine info`
via CLI which is very slow. Works fine on macOS (a bit faster), and very
fast on Linux (since podman machine wasn't needed).

I did the following:

* Cache `checkPrereqs()` result on mount and reuse it in `validate()`,
  since podman version and rootful status don't change during a session

* Remove duplicate `checkPrereqs()` call on mount (was called in
  `onMount` and again inside `validate()`)

* Add 500ms debounce so it triggers after it's all typed in.

* Call getMachineInfo once in prereqs, rather than multiple times.

### Screenshot / video of UI

N/A, no UI changes.

Fixes #1701

### How to test this PR?

1. Run on Windows (or macOS.. but error isn't hit as often)
2. Go to Build Disk image page
3. Confirm no timeout issue
4. Type path in the output folder and don't see any errors / validation
   issues..

There's a few tests covering this as well!

### DIFF
--- a/packages/frontend/src/Build.svelte
+++ b/packages/frontend/src/Build.svelte
@@ -56,6 +56,13 @@ let bootcAvailableImages: ImageInfo[] = [];
 let buildErrorMessage = '';
 let errorFormValidation: string | undefined = undefined;
 
+// Cache the prereqs result so we don't call podman CLI on every form change
+let cachedPrereqs: string | undefined = undefined;
+let prereqsChecked = false;
+
+// Debounce timer for reactive validation
+let validateDebounceTimer: ReturnType<typeof setTimeout> | undefined;
+
 // Specific to root filesystem selection
 // SPECIFICALLY fedora, where we **need** to select the filesystem, as it is not auto-selected.
 // this boolean will be set to true if the selected image is Fedora and shown as a warning to the user.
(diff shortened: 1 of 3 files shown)


cdrage:
This should (hopefully!) fix the CI issues now @cbr7 if you are able to test

cbr7:
@cdrage I'll check it tomorrow morning.

packages/frontend/src/Build.svelte
```diff
@@ -56,6 +56,13 @@ let bootcAvailableImages: ImageInfo[] = [];
 let buildErrorMessage = '';
 let errorFormValidation: string | undefined = undefined;
 
+// Cache the prereqs result so we don't call podman CLI on every form change
```
simonrey1, line 59: remark: usually we are asked to update those Svelte 4 components before modifying them

cdrage: if I am understanding your question correctly. right now we use {#snippet}, {@render} from svelte 5, but rest of the file is still svelte 4?

in the future we'll migrate it completely to svelte 5

Issue #2374, opened 2026-03-16
E2E tests should validate and check overwrite checkbox if it gets displayed during automated test runs


cbr7:
### Is your feature related to a problem? Please describe

Currently occasionally it can happen that the tests try to use a folder that already exists, in which case an overwrite checkbox is displayed on the build disk image page, it would be good for the test framework to check this checkbox if it's displayed before starting the image build.

### Describe the solution you'd like

Check for checkbox existence and check it before building the image.

PR #2389, opened 2026-03-20
chore: check overwrite checkbox in e2e tests

danivilla9:
### What does this PR do?
Checks the 'Overwrite existing build' checkbox in the E2E tests when a folder overwrite is attempted

#2374 

### How to test this PR?
Run the E2E tests

### DIFF
--- a/tests/playwright/src/model/bootc-page.ts
+++ b/tests/playwright/src/model/bootc-page.ts
@@ -151,7 +151,7 @@ export class BootcPage {
         throw new Error(`Unknown architecture: ${architecture}`);
     }
 
-    // Only visible when building on a folder that has a disk image alreadyç
+    // Only visible when building on a folder that has a disk image already
     if (await this.overwriteBuildCheckbox.isVisible()) {
       await this.checkCheckbox(this.overwriteBuildCheckbox);
     }


tests/playwright/src/model/bootc-page.ts
```diff
@@ -149,6 +151,13 @@ export class BootcPage {
         throw new Error(`Unknown architecture: ${architecture}`);
     }
 
+    // Only visible when building on a folder that has a disk image alreadyç
+    await this.overwriteBuildCheckbox.waitFor({ state: 'visible', timeout: 5_000 }).catch(() => {});
```
cbr7: This line isn't really required I don't think, the box becomes visible the second the path is entered, there's no need to wait for it like this, it either exists or it doesn't.

tests/playwright/src/model/bootc-page.ts
```diff
@@ -149,6 +151,13 @@ export class BootcPage {
...
+    if (await this.overwriteBuildCheckbox.isVisible()) {
+      await playExpect(this.overwriteBuildCheckbox).toBeEnabled();
```
cbr7: Also not required, can be removed.

PR #2764, opened 2026-09-15
feat(build): migrate from bootc-image-builder to image-builder-cli

### LINKED ISSUE #2405: Migrate from bootc-image-builder to image-builder
### Epic domain

BIB / bootc-image-builder (https://github.com/osbuild/bootc-image-builder) will eventually be _deprecated_ in favour of the unified `image-builder` CLI (https://github.com/osbuild/image-builder-cli), which will be CLI based, instead of "image-based" (our current implementation). See PR: https://github.com/osbuild/osbuild.github.io/pull/197

We will have to migrate to the `image-builder` CLI which would include multiple changes / "enhanced features":

- macOS/Windows building: Using `image-builder` in the `podman run --privileged` mode (see thread: https://github.com/osbuild/image-builder-cli/issues/428 and documentation here: https://github.com/osbuild/image-builder-cli?tab=readme-ov-file#run-via-container) this still has to be done via `podman run --privileged` since there won't be native macos/windows support via the cli binary.
- Linux: ability to install the `image-builder` binary
- Linux: no longer using `podman run` command in privileged mode and instead using `image-builder` CLI.  (see here how we currently do it with linux: https://github.com/podman-desktop/extension-bootc?tab=readme-ov-file#escalated-privileges-linux this doc would have to be removed too.

### Additional context

_No response_

cdrage:
feat(build): migrate from bootc-image-builder to image-builder-cli

### What does this PR do?

Replaces bootc-image-builder with image-builder-cli
(ghcr.io/osbuild/image-builder-cli).

The old one was deprecated and unused for 4 months.

CLI flag changes:
- Image ref: positional arg -> `--bootc-ref`
- Output: `--output` -> `--output-dir` + `--output-name`
- Type: `--type <t>` -> positional `<image-type>`
- Arch: `--target-arch` -> `--arch`
  (arm64->aarch64, amd64->x86_64)
- Filesystem: `--rootfs` -> `--bootc-default-fs`
- Config: auto-detected -> explicit `--blueprint <path>`
- Chown: `--chown` removed, now a post-build step
- Local: `--local` removed, local storage is default
- Multi-types: one type per invocation only

Build type changes:
- `anaconda-iso` removed (no longer supported)

Changes include:
* Map new arguments from old build image to new one
* Remove anaconda ISO
* Update build page
* Only one output type per build

### Screenshot / video of UI

Screenshot of using "radio" buttons instead of checkboxes for selection:

Screenshot 2026-09-15 at 4 16 08 PM

Closes https://github.com/podman-desktop/extension-bootc/issues/2405

### How to test this PR?

ALL OS's:
* Test each output on arm64 (macOS), Linux (amd64)
  and Windows (amd64).
* Make sure each output completes correctly.


bmahabirbu (approved):
Overall Looks good but I found two concerns. I dont think they're blockers but imo worth looking at

Tested two bootc general containers

quay.io/centos-bootc/centos-bootc:stream10
quay.io/fedora/fedora-bootc:44

**Concern 1**: 

fedora shows "default" grayed out good
centos shows "default" usable good

But

When I have both images pulled the bootc extension defaults to fedora (if that was the last used) for the selection and when I switch to centos the button is still grayed out.

**Testing phase**

Testing: Macos M4 Tahoe 26.6.2 podman 5.8.5

Tested all image types with ARM64 all build properly with fedora and centos. 

Checking configuration logs for a specific build and all selections from the ui are in place

**Concern 2**:
 Builds fail on centos when selecting BTRFS filesystem. Seems the centos image doesnt have that filesystem. Wondering if its possible to gray out the BTRFS if a known image has this problem but that could be a followup

Image

Testing: Windows 11 25H2 AMD Ryzen vis wsl podman 5.7.1

all images built passed with same observations

example of some of the builds on windows

Image

Testing: Linux Fedora 42 (AMD Ryzen) native podman 5.8.2

all images passed build and same observations as before

Had to type my password in each time before build start which makes sense since we need rootful access!

example of some of the images built

Image

cdrage:
Thank you for the review! :)

I opened up the two issues to cover it since it existed before this change:

- https://github.com/podman-desktop/extension-bootc/issues/2782
- https://github.com/podman-desktop/extension-bootc/issues/2781

@bmahabirbu

